PR #77, opened 2015-07-02
Update en_US.lang

Kodos-Atoz:
Correct some misspellings and spacings.

Probably incomplete, but I'm tired and cba to go over every word.


xbony2:
grag

Kodos-Atoz:
uwot

PR #126, opened 2015-08-08
Fixes to Large STEAM Turbine output

pyure:
* Using multiple input hatches now functions properly
* Flow efficiency now operates purely on variance between optimal and actual flow ((actual + optimal) / optimal) (No longer do 150% overflow)
* Disabled 256-tick averaging
* Low outputs during startup or poor efficiency no longer produce 0 output and halt the turbine

### DIFF
--- a/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_LargeTurbine_Steam.java
+++ b/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_LargeTurbine_Steam.java
@@ -66,21 +66,29 @@ public class GT_MetaTileEntity_LargeTurbine_Steam extends GT_MetaTileEntity_Larg
 		
 		@Override
 		int fluidIntoPower(ArrayList<FluidStack> aFluids, int aOptFlow,	int aBaseEff) {
-		    int tEU=0;
-		    int tOut=0;
+    		    int tEU=0;
+		    int totalFlow = 0;
+		    int flow = 0;
+		    
 		    for(int i=0;i<aFluids.size();i++){
 		    	if(aFluids.get(i).getFluid().getUnlocalizedName(aFluids.get(i)).equals("fluid.steam")||aFluids.get(i).getFluid().getUnlocalizedName(aFluids.get(i)).equals("ic2.fluidSteam")){
-		    		tOut = Math.min((int)(aOptFlow*1.5f),aFluids.get(i).amount);
-		    		depleteInput(new FluidStack(aFluids.get(i), tOut));
+		    		flow = aFluids.get(i).amount;  
+		    		depleteInput(new FluidStack(aFluids.get(i), flow));
+		    		totalFlow += flow;
 		    	}
 		    }
-		    tOut = getAverage(tOut);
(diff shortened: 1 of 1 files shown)


Blood-Asp:
From how i understand the turbine does now instead of max 150% simply uses everything that is in the hatch. 
depleteInput amount = how much is in there.
small reminder: depleteInput does deplete all hatches having the same fluid, the loop is only in case different fluids are there.
That leads to the efficiency problem. Consume 200% = 200% efficiency, so supplying more than the optimal flow now rises the efficiency.
Also, what happens when 0 fluid is consumed? Math.max(1, tEU) limits the minimum output to 1, so the turbine never stops.
Is there a special reason you disabled the 256 tick average?

pyure:
Ok now that i've gotten some sleep:

1) the variance works in both ways.  200% flow = INEFFICIENCY, not efficiency.  Note that i take the absolute value of their difference.  ( tested, confirmed)
2) I agree with you about restoring the 150% consumption, because I don't like voiding 40k L of steam just to start the turbine with a good hatch ( tested, confirmed)
3) If 0 fluid is consumed, it doesn't do Max(1,X) ( tested, confirmed)
4) the 256 tick average thing doesn't seem to be necessary anymore now that the math works. You want it back?

With your blessing I'll restore 150% consumption (and the 256 tick average if you think it adds something).
150% consumption means that no matter how much you over-fill the steam, it will always run at at least 33% efficiency (which is fine imo)

pyure:
Regarding zero fluid: the MAX clause only happens if totalFlow > 0.

Regarding the depleteInput hitting all hatches: good to know.  Should I change anything there?

Blood-Asp:
1. Hmm, right. I forgot that Math.abs ignores plus/minus.
2. Thats one of the reasons. The other is, that lazy players have the possibility to simply not regulate the flow and still have 66% efficiency. How about max 125% flow, so 83% efficiency without any regulation? That would make the large turbine more efficient that single block turbines with good material turbines used. So using the optimal flow would only be a bonus for those that want to reach the max.
3. ok
4. In case pipe sloshing ends up creating a unregular steam supply, the efficiency would lower without the player having a good way to counter. The averaging would stabilize output and efficiency without using that much processing power.

5.(depleteInput) Currently it makes no difference. But when limiting to 150% you have to look out for it.

pyure:
2) I'll tweak that
4) I was afraid you'd say that but I think you're right.  The averaging makes the turbine change its speed more realistically/gradually, and minute changes to flow won't have a big impact.
5) Ok I'm walking through deleteInput trying to wrap my brain around what's actually going on here.  I'll make the necessary changes.  

I'll kill this pull request if I can figure out how.

pyure:
"small reminder: depleteInput does deplete all hatches having the same fluid, the loop is only in case different fluids are there."

Blood, if i read this right, depleteInput escapes out of the first hatch if it successfully depletes input.

Blood-Asp:
It only reaches the return in the loop, once the drained fluid is equal or more than the requested fluid. Or at last that is how it should be.

pyure:
I couldn't wrap my brain around it, so I simply tracked requested outside of the depleteInput call.  Works fine.

I'm just doing another round of testing before initiating another pull.

PR #127, opened 2015-08-09
Update Large Steam Turbine

pyure:
* Multiple input hatches now handled properly.
* Now use 125% maximum input consumption and penalty
* We now track averageFlow and totalFlow separately: we don't want turbine byproducts generated from the average, but the actual consumption.
* We track 125% allowance via remainingFlow.  We do still attempt to pull 125% from each hatch, but only up to our 125% overall allowance. (So even with two hatches, we can still only pull 500 steam total from a 400 steam rotor)
* With remainingFlow, we gain optimization: we don't even check unnecessary hatches if the first one provided enough steam.
* Reinstated averageFlow.
* added a ton of comments for your convenience.


pyure:
Done this with the HS version as well, only tweaked few lines necessary.  But want your feedback before getting ahead of myself.

Blood-Asp:
Seems ok on the first look.
Will merge and make my own quick look in eclipse.

PR #132, opened 2015-08-10
Fix to LHE in/out ratio (1:160)

pyure:
Blood, I'm submitting this just for you to look at.  
I think your useWater method should work just fine but I've had no luck with it.  See if this is an acceptable workaround.

As discussed, the method now accepts the "target steam output".  We then determine how much distilled water we're going to consume (to the nearest integer) and calculate the "actual steam output" from that.

Testing demonstrates virtually identical overall steam output to the old method.  The big difference is that we always get a perfect 1:160 ratio of distilledwater:steam.

This does NOT investigate whether total output numbers are correct ( I don't know if they are or not ).

Sorry about...
{
  how I do brackets;
}


pyure:
I need to double-check something.  I used useWater where I didn't intend to.  And I want to see if I can fully remove the round() in favour of useWater, because I like blood's useWater idea.

PR #136, opened 2015-08-12
Turbine fixes

pyure:
* removed getAverage() requirement in favour of gradual (maximum) power fluctuations.
* Fixed math in Plasma/Gas turbines, similar to Steam.
* Removed multi-hatch exploits in Plasma/Gas turbines
* Removed problems when multiple, different fluid types passed in simultaneously (Hydrogen + Methane, etc)
* added 125% maximum overflow penalty


pyure:
:)  (Yes I've been watching this thread non-stop)

Blood-Asp:
Still so many other small fixed i'm working on. So release will take 1-2 days.

pyure:
That's why I've held off more pull requests :)

PR #137, opened 2015-08-13
Let GT AdvCircuits trigger the "stepforward" achievement

doushio9967:
..or else anyone that is unifying their circuits to the nicer-looking GT variants will miss out on this achievement.


Blood-Asp:
I will change this (and others) to oredict detect.

PR #140, opened 2015-08-13
Fixes the most terrible bug ever

xbony2:
Since us Linux and OS X users are left in the dark, I came up with the most ingenious fix :P 

(This PR may be a joke, but who knows?)


Blood-Asp:
I expect every linux user should be able to understand that line.

OS X users... well... no comment...

So i will count this for now as a joke. :P

PR #156, opened 2015-08-22
Fixed large steam turbine do not consume steam.

DeckerCHAN:
Large steam turbine do not consume steam as MFR installed. Simply caused by MFR changed regular steam "fluid.steam" to "fluid.mfr.steam.still.name".


Blood-Asp:
the ".still.name" at the end somehow sounds wrong. That should be the unlocalized name of the icon, not of the fluid itself. If it really works this way mfr has some really strange names.

DeckerCHAN:
Indeed, ".still.name" suffix sounds strange, but it works. I've already tested that.
Initially, when I find this bug, I did not realized MFR involved. And I was confused after I read GT code because everything seems good. Then I decide to change GT to output fluid name,"fluid.mfr.steam.still.name" got. Add this name to GT code, everything works fine.

DeckerCHAN:
Or maybe add an unification to unify MFR steam to GT steam.

PR #232, opened 2015-10-15
Removed JD comments

ShawnBuckley:
Removed JD comments


Blood-Asp:
Could you do the same for the experimental repro (just uploaded)? Merging was not really possible anymore with so many differences.

ShawnBuckley:
Sure, I'll take a look at it.  Is it different enough from this repo that it can't be a branch?

PR #248, opened 2015-10-19
Update gradle dependencies

ShawnBuckley:
This change updates gradle to use artifact repositories for all dependencies except for IC2 Nuclear Control.  It still needs to be in libs/ in order to build.


Blood-Asp:
I have no idea how exactly gradle works,  so also no idea what exactly this does change. Is it allright to have forge 1448 in the properties when i actually use forge 1481?

JasonMcRay:
I guess it should be ok to change the version inside build properties to any version you want to use.

ShawnBuckley:
Yeah, changing the versions in build.properties will change what gradle builds it with.

I've grabbed all the newest versions that other mods were offering that did not require any API changes.  Forestry 4.x would have required some code changes so the latest 3.x (3.6.9.28) was used instead.

PR #259, opened 2015-10-21
Simplify IC2 machine recipe crafting disabling

ShawnBuckley:


Blood-Asp:
This disables generator and electrolyzer that were still craftable by default.

PR #449, opened 2016-03-11
Fix for large steam turbines producing only half EU/t

Timeslice42:
Fixed large steam and large HP steam turbines so that they produce the standard 2mb/EU and 1mb/EU, respectively.


Timeslice42:
Hang on, this missed a change

PR #450, opened 2016-03-11
Fix for large steam turbines producing only half EU/t

Timeslice42:
Fixed large steam and large HP steam turbines so that they produce the standard 2mb/EU and 1mb/EU, respectively.


Timeslice42:
Here's hoping you don't ask me to squash this commit, because I always screw that up.

Techlone:
Why do you think that it's a bug?

Timeslice42:
Why do I think that large turbines producing half of the EU/t that they're supposed to is a bug? I'm pretty sure that question answers itself.

All other GT and IC2 power generators generate (or consume) at a rate of 2mb/EU for steam and 1mb/EU for superheated steam (for the ones that accept it) modified by efficiency. The generation rate for steam was established in GT 4.

Blood-Asp:
Tested it after adding Nuclear Control support: The turbines use the correct ratio of 2L Steam per 1 EU or 1L superheated steam per 1 EU.

But found and fixed another bug: Suppliying above the optimal flow lowered the output again but that should stay at 100%. So now oversupply is 25% more efficient than before.

Timeslice42:
Hopefully it's just and issue with the version I'm using, but it was definitely producing half the EU it should. Cheers for Nuclear Control support!

Issue #407, opened 2016-01-05
Suggestion: Hard / Soft Hammer icon change


DarknessShadow:
Right now the icons for both the soft and the hard hammer look exactly the same and the only way to see the difference is by hovering your mouse over them.
<a href="http://imgur.com/qOCNXxb"><img src="http://i.imgur.com/qOCNXxb.png" title="source: imgur.com" /></a>

I think simply rotating the soft hammer icon by 90° would be enough to end all confusion. :)

Timeslice42:
Although renaming it to a mallet and changing the graphic accordingly would be a nice long term goal...

draknyte1:
Closed now ;)

Blood-Asp:
Right...

PR #482, opened 2016-04-10
Changes Soft Hammer -> Soft Mallet & Flips Icon. Makes them easier to differ now.

draknyte1:
Makes the Two hammers distinguishable.

Now it's a soft Mallet with an inverted icon, as requested by @DarknessShadow - https://github.com/Blood-Asp/GT5-Unofficial/issues/407

Conflicts:
-   src/main/java/gregtech/api/enums/OrePrefixes.java
-   src/main/java/gregtech/api/enums/ToolDictNames.java
-   src/main/java/gregtech/api/enums/TextureSet.java
-   src/main/java/gregtech/api/util/GT_ModHandler.java
-   src/main/java/gregtech/common/items/GT_MetaGenerated_Tool_01.java
-   src/main/java/gregtech/common/tools/GT_Tool_SoftHammer.java
-   src/main/java/gregtech/loaders/oreprocessing/ProcessingPipeLarge.java
-   src/main/java/gregtech/loaders/oreprocessing/ProcessingPipeMedium.java
-   src/main/java/gregtech/loaders/oreprocessing/ProcessingPipeSmall.java


draknyte1:
It was built against a slightly older version, but I'm new to PR's so excuse the slight mess of files.

The only files effected can be seen in the single commit that's attached.

draknyte1:
I've fixed the few things related to my src files not being latest version.
Only now makes changes where it needs to! Should now be 100% code insert with a minor refactor.

Blood-Asp:
I'm opposed to renaming also the oredict. The visible name can be changed, but internally it should all stay soft hammer. Otherwise many minetweaker scripts for GT had to be rewritten.
Also, is there a reason for the change in the "processingPipe" classes?

draknyte1:
Processing pipe Was only changed to reflect the factor of craftingToolSoftHammer to craftingToolSoftMallet.

I'll re-do it all, so as not to change the oredict, so the processingPipe classes won't be touched anymore. Anything else you wanted changed? Wasn't sure what textureID's to use in the Texture Enum, but they're easily changed after merge if you've got something better in mind.

draknyte1:
So, I think that's it.
Should be conflict free and ready to merge, afaik.

Blood-Asp:
Should be fine this time.

Issue #162, opened 2015-08-26
Plasma Turbine output


AiwendilCiunio:
Currently plasma turbine doesn't output anything, which doesn't really make sense. Helium plasma should become helium, oxygen plasma oxygen etc.

pyure:
Aiwen, have you done research on this?  

The whole notion of a "plasma turbine" itself is probably a bit odd, so trying to determine a sensible byproduct (such as helium) and in what quantity would likely be difficult. 

But if there's a sensible byproduct based on actual research, adding it in is simple.

(tl/dr: I don't wanna code in 1:1 plasma:helium just cuz it sounds cool)

AiwendilCiunio:
No I haven't, actually. But something should come out - plasma is matter after all. I'm not sure what happens to it when it's cooled down - does it break into its components or does it become a normal state of the element (i.e. helium plasma -> helium gas).

We would have to ask some physicist I'm afraid :)

AiwendilCiunio:
I found this though:
http://www.crossfirefusion.com/spacedrive-electric-propulsion/phase-shift-plasma-turbine.html
It's difficult to understand for me though, and I don't wanna spend an hour with a translator :)

pyure:
https://www.iter.org/sci/fusionfuels

Last paragraph suggests that helium is a legitimate byproduct of a helium fusion reaction.  Next, just need to figure out what proportion is returned as inert helium as opposed to converted into energy.

pyure:
I can't seem to find any numbers on this.  The prevailing theme is "we're not certain because we haven't done it in production yet."

Two questions:
1) Do all our plasma types (Oxygen, Helium, etc) have a valid non-plasma-form we can use as a byproduct?
2) Any issues with us arbitrarily picking a percentage...say 50%...that would be ejected as a byproduct?

So if you react 50mb of A with 50mb of B, you'd get (1) Energy + (49)AbstractedJunkAshStuffWeDontDealWith + (50)Helium = (100 total mb)

Taking some liberties with numbers here:  I suspect more helium than this is generated.  But GamePlayBalanceFirst, and 50% is a nice round number

Also, I wonder why we don't use plasma to generate power via steam.  It seems to me we should be feeding plasma into the LHE or an equivalent instead of into a turbine directly.

Blood-Asp:
The only reason for far that large plasma generators have no cold fluid output, is that it never had one, back in GT4.

[...]

PR #491, opened 2016-04-17
Turbine rebalance

mura3277:
Added the Diesel Generator
Added 3 new turbine casings
Changed the different turbine multi blocks to require tiered casings:
Steam = Turbine Casing
HP Steam = Stainless Steel Turbine Casing
Gas = Stainless Steel Turbine Casing
Diesel = Titanium Turbine Casing
Plasma = Tungstensteel Turbine Casing

A couple things to highlight:
This is my first time tinkering with GT multiblock code so it's quite possible I have overlooked/missed something but it seems to be working fine. I really did think that the better turbines were too easy to make, especially the plasma turbine. I hope you agree with my changes.

For the Diesel Generator, all I did was copy and Plasma Generator class and instead of having optimal flow \* 40 it's \* 5. In my opinion, the Diesel Generator should be the 2nd best Large Turbine so I thought it was an acceptable amount of energy production. Feel free to make changes.

Let me know if there is anything I'm missing or need to fix.

Issues I know of:
I couldn't figure out how to get the input hatch, maintenance hatch and dynamo to match the casing texture.


MauveCloud:
I like the idea of having a multiblock diesel generator, but I'm not sure it makes sense for it to be so similar to the other large turbines, since that presumably means it will require a metatool turbine item in the interface.  IRL, from what I can find about diesel engines, they require high compression in pistons, and probably wouldn't work as turbines (technically, I found a YouTube video that claimed to be of a diesel turbine, but it looked like it required constant tweaking and didn't seem very practical).  Also, the single-block diesel generators in GT use electric pistons and gears in the recipes, rather than rotors like the single-block steam and gas turbines.

leagris:
@MauveCloud Diesel Turbine generators accept a variety of fuels and are a real thing in use within industrial facilities or in data centers. These are also used in large diesel electric vehicles: (M1 Abraham Tank, Locomotives).
These accept a variety of fuels including gas.

@Muramasa- I really think you should have added diesel generator fuels to the existing Gas turbines instead of adding a distinct type of multi-block. Gas and Diesel turbines are the very same thing.

MauveCloud:
@leagris sorry, I'd forgotten that the M1 Abrams engine was actually a turbine.  However, the fuels listed for it are all in liquid phase at room temperature and standard pressure afaik, so I'm not so sure a diesel turbine is really the same as a gas turbine (which takes fuels that are already in gas phase at room temperature, like methane and hydrogen)

Blood-Asp:
As almost all huge generators for emergency power or huge vehicles i know of use piston engines, so i took some time to look at wikipedia why that is so. It seems also most huge fuel based power plants i could find use the fuel to heat water for steam turbines.

As far i could see it, the reason for that is that turbines are better fit for light fuels and gasses, while the efficiency of them goes far down the more heavy the fuel is.
For example the mentioned Abrams turbine engine: It is quite compact and has a good power curve, but the fuel efficiency of it is really bad. About half the efficiency of other tanks with the same engine power. Also the high rotation speed makes complex gears nessary.
So a fuel turbine for power generation seems to be a bad idea. 

So back to the PR: The changes&additions are quite big, so it will take some time to look trough it.

leagris:
@Blood-Asp Good point about efficiency. Though, pushing further this reasoning, it also question the pertinence of Plasma turbines as already mentioned here: https://github.com/Blood-Asp/GT5-Unofficial/issues/162

mura3277:
I appreciate everyone's feedback on this pull request!

So to explain a couple things, the reason I wanted a diesel turbine is because I really really liked all of the development into Oil/Fuel, but felt like there wasn't a great way to produce power from it other than the single block generators. I did think about adding them to the Gas Turbine @leagris, however the effort put into producing enough fuel wasn't really going to be worth the energy output. I do really feel like the refined fuel category should really generate more power than the others except for Plasma of course.

Making the turbines have tiered casings was because it really was not difficult at all to go from a Steam Turbine to a Plasma Turbine as it just required the main block to be made. I feel like this change balances the turbines a little more, but of course everybody is welcome to offer their input.

@MauveCloud, I agree that it would be more interesting to have a different multiblock to generate power from refined fuels, but that's not really something I could personally do and would come down to BloodAsp, but since he already has  a lot to do, my thoughts were this was the next best thing.

@Blood-Asp, Keep in mind most of the 117 file changes are textures for the different turbine casings.

mura3277:
So I have been messing around with creating a new multiblock for a Diesel Engine and it's going well.

What I'll do is close this PR and submit one that just has the turbine casing changes, then submit the diesel engine in a new PR.

Issue #439, opened 2016-02-18
Flaky behavior of GT axe durability when harvesting brittle charcoal


MauveCloud:
StragaSevera mentioned it with regard to flint axes in the GT5u thread, but I've also seen a similar thing happen with a GT iron axe - the durability bar shows it goes down, then comes back up.  Once the durability is below about 2600 (not certain of exact threshold), the axe breaks for about half a second, but then reappears with durability only down by 25.

Timeslice42:
Sounds like a desync between server and client.

MauveCloud:
Two problems with that theory:
1. I was getting it in single-player (I know it's technically running as a server on the same machine as the client, but that should make it highly resistant to such desync issues)
2. It happened repeatedly as I went through the brittle charcoal - with a diamond GT axe, it started temporarily breaking at about durability 5800 (and with InvTweaks "sort on pickup" enabled I often had to open my inventory to transfer it back to the hotbar)

Timeslice42:
In modded minecraft client/server desyncs rarely have anything to with network traffic (or multiplayer). The client displays on screen what it thinks happened based on your actions. At some point in the (near) future the server responds with what actually happened and the client updates accordingly. Normally this is all without the player's knowledge, but on occasion what the client thinks happened is not what the server says happened, and the server is always authoritative.

There are even a few ways you can see this in vanilla minecraft SSP. When trying to place a block on the cieling at your maximum reach you will briefly see it get placed, and then it will disappear. This is because the client and the server aren't using the same criteria for block placement (I'm hoping this gets fixed in 1.8+, but it's been a long-standing bug). You can also see this with any tool/weapon that has unbreaking on it, especially if you're using the armor status HUD mod which shows the numerical durability of an item you're using. This is because the server and the client don't agree on when unbreaking prevents damage (as they aren't rolling the same pair of dice, to use a metaphor).

The fact that inventory tweaks is a client side mod is why you'll see it move the tool even when it doesn't actually break, as it sees what you see.

Really the only part of this that @Blood-Asp should probably take a look at is why in the world the client thinks the axe is taking 2000+ durability damage. That, and the fact that the shovel is supposed to be the operational tool for the charcoal pit...

MauveCloud:
> Really the only part of this that @Blood-Asp should probably take a look at is why in the world the client thinks the axe is taking 2000+ durability damage.

[...]

Issue #180, opened 2015-09-14
Output busses are not "smart"


Timeslice42:
There's a bug that's annoyed me for a long time, but I never remember to report it: the output bus on multiblocks isn't "smart". For example if I put 64 iron ore and 64 calcite into an EBF, I end up with 64 small dark ash, and three stacks of 63 iron ingots. The last three iron gets vaporized because it couldn't find a slot with 3 free spots, despite the fact that there are three slots with 1 free spot.

willis936:
You need to put a conveyor belt on the bus to move the output to a chest.

DarknessShadow:
>You need to put a conveyor belt on the bus to move the output to a chest.

Either I completely missunderstood @Timeslice42 problem or @willis936 "solution" doesn't have anything to do with the problem

willis936:
My understanding is the problem is that when the output bus is full items get lost.  That is the intended behavior.  This sounds like a fringe case of getting burned by the intended behavior and strictly speaking this case isn't intended.  A conveyor belt would fix his problem.

PaladinOne:
A conveyor belt would *work around* his problem. The actual problem is that the busses only check for space in single slots, and if they can't find the space in any slot, they void the entire output stack, rather than placing as many as possible or spreading them across multiple output slots.

Timeslice42:
And PaladinOne nails it in one.

Blood-Asp:
Really minor improvement that needs quite a bit of coding. Will get a really low priority... So as long none does a pull request, it will likely stay as it is now.

PR #499, opened 2016-04-28
Improved output bus handling

MauveCloud:
Should address bug #180 (Output busses are not "smart")


MauveCloud:
The change to GT_Block_Reinforced.java should fix #439, as well as change the preferred tool to a shovel instead of an axe.

Blood-Asp:
Does it also help with not working in advanture mode?
Also i would prefer to stay with an axe. More intuitive.

MauveCloud:
How is an axe intuitive?  As mentioned in the bug, other mods that add charcoal pits use shovels.  It doesn't help with not working in adventure mode.

Blood-Asp:
Shovel in TFC or what other charcoal pit? For me personally i would only think of trying axe or pickaxe to break in coal transformed wood blocks.

MauveCloud:
I don't know what other mods add charcoal pits/piles, but Timeslice42 said that real-life charcoal piles also involve using a shovel to get the charcoal when it's done, and the picture "Opening of the charcoal pile" at https://en.wikipedia.org/wiki/Charcoal_pile seems to support that (I admit I don't have personal experience with charcoal piles in real life).

Edit: I found a couple of YouTube videos demonstrating how to make charcoal with a charcoal pile.  One video had a guy just using his bare hands, another had a group using rakes and shovels.

Timeslice42:
You are now my #1 person of the month, Mauve. Thanks for both of these :)
I knew fixing the output bus wouldn't be as hard as Blood thought...

PR #500, opened 2016-04-28
Large Diesel Generator

mura3277:
So the multiblock still isn't done, but I'm submitting the PR now to open it up to a discussion.

I'm not really sure how much power it should produce or what the best way would be to calculate the output. It's probably best that this is done by someone who already knows how the current diesel generators work.

How I see the diesel multiblock fitting with the rest is having the 2nd best power output (1st would be plasma turbines). So it would be interesting to have it require more "maintenance" (Reason for lubricant and possibly coolant).

The current todo/ideas is as follows:
~~Require Muffler top centered (Currently anywhere)~~
~~Require Titanium Pipe Casing in the middle of the multiblock~~
~~Implement output calculation~~
~~Implement coolant requirement (Water or IC2Coolant?). Currently requires lubricant.~~
~~Possibly add different tiers of large diesel generator?~~
Implement Achievement
~~Finalize output~~

Edit:
I forgot to mention that the multiblock is pretty much the same as the Large Heat Exchanger except that it's only 3x3x3 instead of 3x3x4.


Blood-Asp:
Personally i would say two power versions. 2kEU/t and 8kEU/t. Efficiency 150% but needs as you said lubricant and/or water. Maybe default mode is 2kEU/t and 1L lubticant every 6 sec and adding water it speeds up to 8kEU/t, needing 1L lubicant every sec (a bit inefficiency) and 10L water per tick.

leagris:
Diesel engines use little lubricant as top engine parts (valve train, push rods...) are lubricated by Diesel fuel itself. Lower part on heavier load (crankshaft, camshafts) need dedicated lubricant circulated with pump and collected/stored in oil pan. Lubricant live long in a Diesel engine. It is not consumed but contaminated and oxidized despite circulating through an oil filter. I don't think oil would need frequent changes. Though, Gregtech 3 used to provide jerry-cans. Those filled with fresh lubricant would be a good consumable to replace turbine.
Now you could also fill an internal tank with fresh oil and need to dispose of contaminated used oil. Maybe something worth distilling into tar concrete and some lighter oil by-products.

About cooling, Diesel engines runs hotter than Gas engines. They use less cooling capacity power-wise. As well, coolant (water kills engines cooling circuit), circulate and recycle in a closed loop. Unless tubings are damaged, you don't need to replace coolant that often.

Now lets consider your "Large Diesel" will accept other light fuels and it change rules about lubrification, cooling.
Ethanol(bio) is a killer. It harden seals, capture water from ambient air, tend to degrade lubricants faster.
NitroFuel is like an overclocker. More HP, more heat in the same space. Require more beefy cooling.

These are more directions than hard requests here. But there are probably things to exploit with fuels types, cooling and lubrication requirements. That would serve as consumables items in place of turbines. You could even imagine adding blower turbo block to boost EU/t at the cost of some efficiency loss, more cooling requirements... Or even direct oxygen intake, giga boost with that ;D

mura3277:
A couple questions @leagris 
Would water be a good candidate for coolant?
What could the output multi be for supplying Oxygen gas?
Would it be better to "consume" lubricant but output "waste"?
What would be the best liquid for the "oil"?

So for the base lube/coolant/oxygen usage:
Lube: 1L/6s
Coolant: 10L/6s \* (fuelValue / 128)
Oxygen(if supplied): ?L/6s

Thanks for commenting!

leagris:
@Muramasa- 

>  Would water be a good candidate for coolant?
> Re-use IC2 Coolant, it is here already. Multi-block diesel should include a coolant block (radiator with fan), an exhaust block with pollution when implemented.
> 
> What could the output multi be for supplying Oxygen gas?
> Output? I think you are talking about intake. Air intake block. Need free air block or a supply of oxygen. If obstructed, diesel engine can't work. If provided oxygen, well cool but beware of fast overheat and damages.
> 
> Would it be better to "consume" lubricant but output "waste"?
> I'd say output waste oily fluid you have to pump out/open lubricant maintenance hatch, insert empty bucket/fluid cell, get filled with waste oil, insert new lubricant cell (oil change)
> 
> What would be the best liquid for the "oil"?
> Existing Lubricant fluid is close to what's used in engines, so lets use it. Engine lubricant is made of oil synthetic or mineral (petrol) + detergent + thickness/heat grades regulators. GT Existing lubricant is mad of oil (seedoil) + detergent (soapstone)
> 
> So for the base lube/coolant/oxygen usage:
> Lube: 1L/6s
> As mentioned above, lubricant change (disposal + recharge) should be on player interaction only.
> So make it not so frequent. 1000L should last a realtime week of constant use.
> 
> Coolant: 10L/6s \* (fuelValue / 128)
> Coolant is lost by evaporation in normal use. The more it heat-up, the more you loose.
> So indeed dependant on fuelValue × turbo overclock % (blower, oxygen intake, nitrofuel)
> Lets say you have radiator vent blocks. Normal rate require only one radiator block.
> Radiator block needs free space in front or above (depend where your place it)
> Radiator block has an internal tank for X Litters of coolant. Higher tier radiators contains more coolant, are able to do more heat dissipation.
> During operation, coolant is lost function of power output rate% (1L × power%) / 720 ticks so consume 100L / hour at normal engine operation rate. If boosted, it loose coolant faster.
> That is coolant lost = Nominal 1L × 3^power%/72ticks. At 130% power it consume 400L / h
> With a single basic tier 1 radiator block you have a max tank capacity of 8000L, That leave you enough coolant to operate 20 hours without refill if your run it at normal 100% rate (no boost)
> You may choose to add higher tier (higher capacity) and more radiator blocks to last longer without coolant refill, or have a contant pipe-in supply of coolant (cold). No hot coolant is returned. This a closed loop cooling system. But if you need to pipe-in coolant you need another distinct input hatch because radiator blocks need free space.
> 
> Oxygen(if supplied): ?L/6s
> Oxygen consumption is directly proportional to fuel value. If you venture into that route. You need 1L of oxygen per tick and per fuel value/10000, So to satisfy combustion of a 128000 fuel value, you'd need 115.2L of oxygen/tick, or 2304L of oxygen /s
> Oxygen boost poer% to 200% so ⇒ fuelvalue/10000 × 3^2

About structure and design. Radiator blocks should replace the turbine face to look like grille vents.
Maybe 1 to 8 radiator blocks.:
1 controller front center
1-8 radiator blocks front
1 fuel input hatch middle side
1 optional oxygen input hatch middle side
or 1 air intake middle side (vent)
1 coolant input hatch middle side
1 maintenance hatch middle side
1 oil gauge and filler trap middle side
1 waste oil output hatch middle bottom
1 exhaust block middle upper
1 dynamo hatch center rear

As opposed to Large Turbines, the Large Diesel is suitable for moderate on-demand operation.
when started, it consume 10L of fuel immediately and begin power output at start+200ticks (10s) at nominal rate. Stop is immediate.

If running out of coolant, it accumulate defects until it stop operating (overheat protected)

All other as diesel casings.

Blood-Asp:
For all the output calculation and stuff, remember that first comes balancing. Realism is always less important.
Calculate how expensive it is to produce certain stuff(lubricant, oxygen, whatever) and balance used amount against that, not how much it would use in real.
The idea of giving different fuels different characteristics (more or less lubricant or whatever)  is something i'm opposed to. Not a good idea balancing and coding wise. All refined fuels only should differ in fuel value per L.

leagris:
Lubricant and Coolant can be ignored.
Air intake with free block in front should be mandatory as well as exhaust.
A Turbo intake could work with small turbine rotors and be required for higher tier Large Diesel.
Direct oxygen has no real life equivalent. That's pure fantasy, so no need to seek realism.
Now after playing GT5 since months, I find I have plenty of unused oxygen stored from all the ore processing. More than I have use for. Using oxygen to boost Large Gas and Large Diesel Generators is an option to sink that extra oxygen to something useful.

mura3277:
Ok well first we should probably decide how the multiblock should be made. I like @leagris's idea of having a "better" Diesel Engine that can use oxygen to boost the output. Would a "higher pressure" version be the right description?

So Large Diesel Engine:
Titanium Casings, Dynamo, Muffler, Air intake and however many input hatches needed.

Large HP Diesel Engine:
Tungstensteel Casings + All Components Normal Engine, must be supplied with Oxygen gas and requires a "Turbo" or "Forced Induction" block on one of the casings.

If this looks good, we could then have a static Lubricant, Coolant & Oxygen requirement for the Normal and HP Engine and wouldn't be determined by a fuel value.

I also really like the idea of having to put a small turbine rotor in the "Forced Induction" block. I'm not sure if I'll personally be able to do that, but I'll have a look once everything else is implemented.

Blood-Asp:
How the multiblock is made does not change much balancing wise, so not much i can complain there. But i would not call it HP Diesel, but Turbo Diesel Engine....

leagris:
Think there few modern large diesel generators without turbo for forced air intake. But lets consider first tier of large diesel does not have a Turbo compressor fan.

Scale is a reason for overall increased fuel efficiency, at the cost of start-up cost where NO EU is output intil diesel is Up-to-speed and in sync with the power grid. (EU/t generated by generator match EU/t for dynamo hatch.

Forced Air intake would increased EU output rate. Does not increase fuel value but increase burn rate.
So Tier1 Large Diesel would only be capable of EV 2048 EU/t in its Tier1 and IV in Tier2 with Turbofan

Lubricant, Coolant, Oxygen not sure it is appropriate.

If you consider Fuel values of Diesel, Bioethanol or even NitroDiesel, a Large Diesel would suck fuel like a greedy beast to output even 2048EU/t with 128000EU per 1000L Diesel @ 100% efficiency. (100L or an IC2 cell every 3s). Low fuel values of in-game fuels are things to consider about these large diesel and it is part of balancing considerations.

MauveCloud:
> But i would not call it HP Diesel, but Turbo Diesel Engine....

One problem I see with that: the word "Turbo" is already used for all three single-block HV-tier generators (steam, gas, and diesel), as well as a couple of single-block processing machines (the HV-tier centrifuge and the EV-tier fluid canner).

mura3277:
So sorry for not being active with this PR, I had a crazy month.

Anyway so after reviewing things the differences between the normal Diesel Engine and the "Turbo" version, I'll probably just combine the two multiblocks to simplify things since the only difference was one had to have a "Turbo Hatch". Unless of course somebody can think of interesting ways to differentiate them, I'll probably just have the "Turbo Hatch" optimal.

Now onto the multiblock changes:
image

It is now 3x3x4 (Like the turbine) and requires the front 8 blocks to be a new "Engine Intake" casing(They require air blocks in front of them!, temp texture), a muffler on the "top middle back", a maintenance hatch and finally the two hollow spaces in the middle need to be titanium gearbox casings. And of course optimal additions are an Input hatch for fuel, the Turbo Hatch (Probably a temp name) and a Dynamo.

Engine Output:
So it now outputs either a max of 2048 or 8192(With turbo hatch and turbine. requiring the turbine item and damage is not implemented yet however.) however if the "startup %" is below or equal to 20% if does not output power(@leagris suggestion). If it's over 20% then it will output EV or IV depending on setup(Multiplied with current startup %).

Fuel Consumption:
Based on the Diesel fuel value(@128,000EU, but in code is 128), the consumed L/t is the baseEU(Either EV or IV) / Fuel Value(Which again, is 128). So for Diesel at EV it would be 16L/t and 64L/t at IV.

Final points:
I personally really like the idea of having more "maintenance" by requiring the multiblock to require something like lubricant to run efficiently. But if anyone can think of more interesting ways to make it better then let me know.

Blood-Asp:
I did lose the overview a bit. Is the implementation here now done and ready to merge or still work to do?

mura3277:
There is still a few things to work out:
~~I couldn't figure out how to get the turbo hatch (copied from maintenance) to accept items in its slot~~ Fixed
~~Decide if lubricant should still be required for the multiblock to run~~ Decided to implement it
~~Check and damage turbine appropriately. Possibly the same as turbine damage? (0.2 of EU in damage over 3000 ticks)~~

mura3277:
So I am leaning towards using Oxygen gas to boost output rather than a small turbine. I could possibly allow either route however.

@leagris thoughts?

leagris:
As long as oxygen requirement does not impair net efficiency, should be good.
Remember:
Light-Fuel is 256000EU/1000l
Nitro-Diesel is 512000EU/1000l
At 2048 EU/t that's respectively 125 and 250 ticks per 1000l fuel at 100% efficiency.
1/4 of that at 8192 EU/t: 31.25 ticks and 62.5 ticks
Extracting that much Oil, processing it into Light-Fuel or Nitro-Diesel has already a significant energy cost.
Add cost of producing oxygen and that Large Diesel risk being not worth it.

Would it be possible to have the Large Diesel output 4 Amperes at 2048 EU/t when operating with oxygen input and 1 Ampere without?
Like structure could accept up to four dynamo hatches for 4×1Ampere=4ampères.

mura3277:
So currently the Oxygen consumption /t equals the fuel consumption /t. Maybe it should be reduced to a 1/4 of that? So with Nitro, the consumption/t would be 4L(2048Out) so Oxygen consumption would then be 1L/t.

I'm not sure if I can set the amps without having more dynamos. Adding more dynamos would mean a structure rewrite, what would be the advantage of 2048@4Amp over 8192@1Amp?

leagris:
> what would be the advantage of 2048@4Amp over 8192@1Amp?
1. Keep same distribution cabling/transformers if you run with or without Oxygen boost as EU/t would not change.
2. Keep the same fuel consumption rate with or without oxygen. Means you run at 400% efficiency with oxygen = Very good incentive to run your large diesel on Fuel+Oxygen even with the extra cost of processing oxygen.

Well, if that make tpo much work, just do the 8192@1A with oxygen. One can still hook an IV Dynamo Hatch and down transform to EV. Same result = 2048@4amp sustained with Oxygen and 1 out of 4 ticks without. But please make oxygen boost not increase fuel consumption rate (keep 400% efficiency).

@Blood-Asp one of our player noticed Single-block diesel generators have non-constant EU output when using Nitro-Diesel. He had to connect a battery buffer to recover a sustained constant output. Maybe this could be fixed.

mura3277:
I probably didn't explain that very well, but oxygen won't the change fuel consumption.

What I was saying is the Oxygen usage/t = Fuel consumption a tick(Doesn't change it). However I'm thinking it should probably be 1/4 of that.

So if you're using Nitro Diesel, Consumption =  16mb/t for Nitro diesel and 4mb/t for oxygen.

Edit: A transformer would definitely be the easiest solution. If there is a way to change amperage without adding more dynamos then I'll change it to 4 amps of 2048.

mura3277:
Swapped the fluidstack outputs of cracked heavy fuel to make more sense when you set up a Oil -> Plastic processing line.

@Blood-Asp This is now ready to merge 

Edit:
It also turns out that Oxygen and Lubricant are both byproducts of the Oil to Plastic processing line which actually works out really well.

Techlone:
Bugs:
1. Need Gear Box Casing instead of Pipe Machine Casing (to change the tooltip).
2. Recipe of Intake Casing is absent.
3. Oxygen and lubricant are consumed even without some fuel when Diesel engine is enabled.
4. In the infotext Fuel Consumption measured by L/**t** but Fuel Value measured by EU/**mb**. Liters or millibuckets - time to choose.

Note.
I think that 100% efficiency without oxygen and 400% efficiency with oxygen isn't correct. I spend the same value of fuel per EU in both cases and eff **is** 100% in both cases. It's just faster.

mura3277:
Fixed all but No. 3 since I can't seem to re-create it. When I try to start the engine with only Oxygen and Lubricant is correctly doesn't run.

Techlone:
Also the gui hint isn't correct. It looks like LHE build.
image

> When I try to start the engine with only Oxygen and Lubricant is correctly doesn't run.

Hmmm. Can't reproduce now. I forgot what I did...

mura3277:
I found out what you did. It happens when there is fuel in the hatch, but not enough for the calculated consumption/t.

mura3277:
@Blood-Asp, What do you think about @Techlone's point?
"Note.
I think that 100% efficiency without oxygen and 400% efficiency with oxygen isn't correct. I spend the same value of fuel per EU in both cases and eff is 100% in both cases. It's just faster."

Should using Oxygen boost the fuel value itself or should just stick with the 400% efficiency increase?

Blood-Asp:
I would likely do something like 3x EU/t by 2x fuel consumption. So 50% rised efficiency and much higher troughput.

natzgzul89:
I read everything you've written.
1. When using the engine power turbine uvelivivaetsya 2 or even 3 times.
2. using nitro fuel unlike the conventional fuel engine power is increased by approximately 4 times
3. The engine has 3 cooling method
 1) air
 2) Liquid
but there is a cooler for lubrication in any of them. 2 can be used immediately one liquid lubricant (which is almost not consumed in operation) and a second fluid (water, distilled water, antifreeze .....)
4. The turbine is often out of order depends on the fuel.
5. the engine without load almost does not consume fuel (0,01 L/t)
Consumption depends on the load:
 if the load generator to the full (8k 4A) the engine will consume the maximum amount of fuel for example 10 L/t
 2k 4A, 8k 1A = 5 L/t
 for 2k 1A = 1.3 L/t.
if the storage batteries connected to the output of the load will be maximum and the engine will consume the maximum amount of fuel until the batteries are charged.
(прочел все что вы написали.
1. при использовании турбины мощность двигателя увеличивается в 2, а то и в 3 раза.
2. при использовании нитро топлива в отличие от обычного топлива мощность двигателя увеличивается примерно раза в 4
3. у двигателя есть 3 способа охлаждения
 1) воздушный
 2) жидкостный
но при любом из них есть радиатор для смазки. можно использовать сразу 2 жидкости 1 смазку (которая почти не расходуется при работе) и вторую жидкость (вода, дистиллированная вода, антифриз.....) 
4. турбина часто выходит из строя зависит от топлива.
5. двигатель без нагрузки почти не потребляет топливо (0,01Л/т)
потребление зависит от нагрузки:
 если нагрузить генератор по полной (8к 4а) то двигатель будет потреблять максимальное количество топлива к примеру 10л/т
 при 2к 4а, 8к 1а = 5л/т
 при 2к 1а = 1.3л/т.
если подключить хранилище батарей к выходу то нагрузка будет максимальной и двигатель будет потреблять максимальное количество топлива пока батареи не будут заряжены.)

bear989:
the tooltip should say that you need 8x intake casings and front maybe to make it a little bit clearer.

Issue #517, opened 2016-05-22
Fluid Filter crash


Nidan:
Rightclicking a fulid filter with another one disconnects you from the server / crashes in sp
Cause seems to be derefencing a Nullpointer in /src/main/java/gregtech/common/covers/GT_Cover_Fluidfilter.java line 42

Issue #501, opened 2016-04-28
Wrong turbine tooltip for gas flow


Techlone:
The tooltip's gas flow is twice less then actual gas flow. I just checked it in my test world.

[Tooltip's flow](https://github.com/Techlone/GT5-Unofficial/blob/experimental/src/main/java/gregtech/api/items/GT_MetaGenerated_Tool.java#L342)
`speedMultiplier * toolSpeed * 25`

[Actual flow](https://github.com/Techlone/GT5-Unofficial/blob/experimental/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_LargeTurbine.java#L133)
`speedMultiplier * toolSpeed * 50`

PR #518, opened 2016-05-23
Crash  Fix / Various Tweaks

### LINKED ISSUE #268: plasma generator NEI not easy to navigate to
If I search uses of the plasma fluid I don't see the plasma generator.  It's only seen when looking at uses of the fluid cells.

### LINKED ISSUE #367: GUI bug all bronze machines
Gregtech 5.09.16
Forge 10.13.4.1558 1.7.10
images:
http://i.imgur.com/8M3hCeA.png?1
http://i.imgur.com/iTYp79e.png?1

### LINKED ISSUE #478: Missplaced achievement "Full Efficiency"
With current achievement tree, you have to first validate "As Dense As Possible" before "Full Efficiency".

But to create As Dense As Possible you need a Fusion Reactor MarkIII because Neutronium fusion require a starter EU of 600000000 EU you can only achieve with a 16 UV Energy Hatches MarkIII Fusion.
So to validate "Full Efficiency" you need to craft another MarkI, MarkII  and MarkIII since you already creafted a Fusion MarkIII to validate "As Dense As Possible"..

I propose you re-order exchange the place of those two achievements.

### LINKED ISSUE #507: Aspects misspelt in enum
This breaks anything that checks such as the magic absorber making it not recognise them, broken aspects are Cognitio, Exanimis, and Praecantatio.

mura3277:
**Changes:**
Overhauled all mulitblock tooltips and gave them all a similar structure.(besides assembly line). I may need a second person to look though them for derps and to see if they make sense to other people.

Added more scanner info for the turbines and added info for the fusion reactors.
Turbine additions: Optimal flow (L/t), Remaining Fuel (L)
Fusion Additions:  EU/t Required, EU Stored/Max EU Stored, Plasma Output (mb/t)

Fixed crash from issue #517 by adding a null  check [here](https://github.com/Muramasa-/GT5-Unofficial/blob/3c32671623dcabcf25fca16b062f7b600963b1d3/src/main/java/gregtech/common/covers/GT_Cover_Fluidfilter.java#L42)

Fixed wrong Gas flow tooltip #501 

Added export gui buttons to bronze and steel machines. Fixes #367 

Fixed #507, renamed 3 aspects to correct spelling

Changed order of "Full Efficiency" and "As Dense As Possible" to make more sense. Fixes #478 

Fixes #268 by only showing the plasma cells in NEI that have a fusion recipe. Makes it a lot easier to search up.


mura3277:
So for the Turbine Optimal flow scanner info, it has occurred to me that my turbine math might be wrong. If someone who knows how to calculate optimal flow (because my calculations are apparently incorrect) then please let me know.

willis936:
I believe optimal flow input is calculated as this:
input: (optimal flow output as listed in NEI) / (out/in fuel ratio), round to the nearest mB/t
output: input \* (out/in fuel ratio)

This sort of breaks down when using large plasma turbines since the input flow rate is only a few mB/t and the rounding may affect the actual output by several percent.

PR #526, opened 2016-06-12
Make processing array handle multiple input stacks at a time

perkinslr:
Also make fluid and item input split between hatches or stacks count toward satisfying a recipe.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_ProcessingArray.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_ProcessingArray.java
@@ -153,7 +153,7 @@ public class GT_MetaTileEntity_ProcessingArray extends GT_MetaTileEntity_MultiBl
 
         ArrayList<FluidStack> tFluidList = getStoredFluids();
         
-        FluidStack[] tFluids = (FluidStack[]) tFluidList.toArray(new FluidStack[rFluidList.size()]); 
+        FluidStack[] tFluids = (FluidStack[]) tFluidList.toArray(new FluidStack[tFluidList.size()]); 
         if (tInputList.size() > 0 || tFluids.length > 0) {
             GT_Recipe tRecipe = map.findRecipe(getBaseMetaTileEntity(), mLastRecipe, false, gregtech.api.enums.GT_Values.V[tTier], tFluids, tInputs);
             if (tRecipe != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_ProcessingArray.java
```diff
@@ -148,34 +148,12 @@ public boolean checkRecipe(ItemStack aStack) {
...
-                    } else {
-                        tFluidList.remove(i--);
-                        break;
-                    }
-                }
-            }
-        }
-        FluidStack[] tFluids = (FluidStack[]) Arrays.copyOfRange(tFluidList.toArray(new FluidStack[tInputList.size()]), 0, 1);
+        
+        FluidStack[] tFluids = (FluidStack[]) tFluidList.toArray(new FluidStack[rFluidList.size()]);
```
Techlone: Please rename rFluidList to tFluidList. Is it correct?

Blood-Asp:
Is this really fully tested? Changes on the core classes of the recipe system is something i really dislike to do. Most the time any change here breaks something else unexpectly. Best to try this change with an endgame world with many machines online and working for a few hours to see if anything stops working.

src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_ProcessingArray.java
perkinslr: Good catch, it's fixed.

perkinslr:
Aye, I can understand that.  It is about as fully tested as I can make it myself.  I tried it out in creative with a variety of machines, measuring carefully tacking input, output, and power used.  I've also had it up and running for about a week in a late-game server (fusion-III) and no one has found any issues.  The processing array happily moves on to secondary tanks and stacks of input, and 1 stack of clay dust + 3 stacks of stone dust get completely consumed in the mixer (with water), instead of leaving 1 clay dust and 3 individual stone dust behind.  

The only unintended change I've observed is when you provide resources which could satisfy multiple recipes.  Before it would only look at 2 input stacks, which would sometimes cause it to switch recipes even when it had the resources needed to continue with the last recipe used.  Now it sees all the input stacks, so it will continue with the current recipe until it is no longer valid, before switching.  

Unfortunately, the changes presented here were slightly intermixed with my rebalance changes, so instead of pushing those changes back to github to make the PR, I just copied and pasted the relevant sections.  The typo is because I'd made a copy of the initial input array so I could double check exactly how it works, and forgot to change the name back.

I just checked out both changes in Idea and it claims I've no further typos.  Considering my other balance changes, it might be best if someone running a vanilla server gave this a test, but those changes _should_ be orthogonal to these.  

Also, if it helps, both of the files changed can be applied independently.  The change to GT_Recipe won't do ProcessingArray much good, since it currently truncates the input to 2 stacks, but everything else will still benefit from it.  Obviously the change to ProcessingArray won't affect anything outside of the ProcessingArray, but would benefit the PA even without GT_Recipe changing.  It would just leave partial stacks behind in the input, similar to how a centrifuge leaves 2 stacks of 4 when stuffed full of redstone, only it could leave more than 2 stacks.  

If you want to see it in action, I can fire up a creative server with a minimal modpack.

PR #533, opened 2016-06-19
Fix adv miner ii exploit

Techlone:


Techlone:
Nevermind. This PR resolve the problem. only this is important

PR #534, opened 2016-06-20
Bug Fixes

mura3277:
Changes:
- Removed commented out old turbine texture code
- Fixed Diesel Engine "Fuel Value" unit to EU/L
- Fixed mistake with Diesel Engine tooltip. Now asks for Gear boxes.
- Forgot to add recipe for Engine Intake.
- Redone the Gas/HPSteam turbine recipes and swapped casings.
- Fixed Oxygen/Lube being consumed if there is diesel in the hatch, but
  not enough for the engine to run
- Updated Engine GUI


mura3277:
@Blood-Asp Also, I don't think my Optimal flow calculation (In mb/t) for being displayed via the scanner is correct. [Example](https://github.com/Blood-Asp/GT5-Unofficial/blob/experimental/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_LargeTurbine_Plasma.java#l94)

Issue #381, opened 2015-12-26
How often Adv.miner 2 produces byproducts?


ShelfOwl:
The Adv.miner 2 produces byproducts each time it seems. Is it intended?
edit: happen In GT 5.09.16

Blood-Asp:
Yes, intended.
The miner gives overall huge bonus. So the 100% chance is intended. The primary output has a 1-5x random multiplier, so one ore is not 2 crushed and 0.1 byproduct, but in average 6 crushed and 1 byproduct.

ShelfOwl:
Hmm I see. But I saw Adv.miner 2 produces 2-8 crushed and 1-4 byproducts, not 5x...
Thanks reply.

PR #536, opened 2016-06-21
AdvMinerII: check that ore block still exists; consider chance output

Techlone:


ShelfOwl:
I heard the Adv.Miner2 has ability to produce subproducts by 100% chance is intended behavior,
differ to actual behavior a bit, though.
See also: #381

@Blood-Asp do you reconsider it? There are 3 calculations at all.
1. 1-4x 2 Crushed + 1-4x 1 subproduct : actual behavior in GT 5.09.21 or below
2. 1-5x 2 Crushed + 1x   1 subproduct : @Blood-Asp mentioned in #381
3. 1-4x 2 Crushed + 1-4x 1x10% chance to subproduct: this PR I assume (hence it seems 90% chance to produce 1 subproduct)

Techlone:
I think that config parameters is the best solution for this issue.
I'm one of authors of IIA modpack. Our community (and we are too) consider that AdvMinerII should take into account chances of byproducts. Are you think differently? See my first sentence.

Blood-Asp:
The main reason i did not add a chance to the byproduct was lazyness... I do not really care about that point.

ShelfOwl:
For me Adv.Miner2 is just an alternative to exploring veins, is not hard in EV+ age.
So I do not care how effecient it will be. Config is good idea.

Issue #540, opened 2016-06-24
Super fuel recipes


Techlone:
Please check recipes of SSF. They are disbalanced
Simple:
![image](https://cloud.githubusercontent.com/assets/17011260/16329197/5280bf92-39fa-11e6-9471-e92fd6ec4d0e.png)
And more complex:
![image](https://cloud.githubusercontent.com/assets/17011260/16329236/985ba388-39fa-11e6-9c47-fb20c552fbe5.png)
But output is the same - one block of SSF

Blood-Asp:
The first recipe should include the Railcraft coal Coke block. I guess you do not have it Installer but the recipe still got loaded even though it has a Null item in it.

Techlone:
He he I see but Railcraft_1.7.10-9.8.0.0 has been installed.

PR #541, opened 2016-06-24
Fix Super fuel recipes

Techlone:
#540


Blood-Asp:
The diff is a bit difficult to read because of the formatting changes, but if i understand it right, you changed "tile.railcraft.cube" into "cube"? That does not fix the issue.

Blood-Asp:
Fixed it myself https://github.com/Blood-Asp/GT5-Unofficial/commit/e21311b919d8f7f2e6ac15afda1b083179684f61 by calling the coke block directly from the railcraft code: "EnumCube.COKE_BLOCK.getItem()". Now whatevery namechange happens to that block will not break the recipe. I still wonder why railcraft uses close to no oredicts.

Techlone:
> That does not fix the issue.

Strange... I tested it with and without Raicraft and Thaumcraft before PR and all work fine.

Also "cube" is used in pyrolyse oven's recipes [this](https://github.com/Blood-Asp/GT5-Unofficial/blob/experimental/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java#L1287).

Techlone:
> I still wonder why railcraft uses close to no oredicts

Are you saw how to realize RC's trails? A pell-mell of metadata and NBT. It was the hell to me when I wrote scripts for them :)

Blood-Asp:
I should fix the pyrooven recipe too. At last there the recipe will simply be removed when failing and not remain half broken.

PR #542, opened 2016-06-24
Silk-touch then pulverize ores in advanced miner II

perkinslr:


Techlone:
You removed all changes from #536 Why?

perkinslr:
Because I made these changes before #536 got merged and I missed seeing it in my review of what's changed.  It wouldn't cleanly apply here, but I just merged what will and reimplemented the non-null check.

Techlone:
okay

Techlone:
Nonono
Null check not there. I think so that player or something else can destroy block after adding chunk coords data and block will be null. You should write null check after [this](https://github.com/perkinslr/GT5-Unofficial/blob/experimental/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_AdvMiner2.java#L139). `tMineBlock != null`

perkinslr:
But that block will never be null.  It might be Blocks.air, and it might not have a TileEntity, but the block itself will never be null.  Unless you're saying that if it is null this pass should automatically move on to the next block in the list... That make sense.

Dream-Master:
Using this code throw out some error.

```
Error:(142, 72) java: cannot find symbol
  symbol:   variable mle
  location: class gregtech.common.tileentities.machines.multi.GT_MetaTileEntity_AdvMiner2
```

https://www.dropbox.com/s/pyivj6sjc16om5d/Screenshot%202016-06-26%2021.40.02.png?dl=0

mura3277:
The issue is mle is not in the same scope as the methods using it.

Dream-Master:
means what? Sorry i am a java noob. How to fix that ?

mura3277:
Compare to see the change https://github.com/Muramasa-/GT5-Unofficial/tree/Temp

Dream-Master:
@Muramasa 
like this when i compare both files ?
https://github.com/GTNewHorizons/GT5-Unofficial/commit/3d9a779242d3611b1d4da5a0c8b218e0708bbf4a

mura3277:
Yep. It's untested of course but it will at least compile. However I can't see there being any issues.

Dream-Master:
Ok thanks

PR #547, opened 2016-06-30
Make plasma turbine output cooled versions of plasma

perkinslr:


ShelfOwl:
Looks some bugs with no description are there.
So I recommend you know:
- There are 3 types of fluids: air, liquid, and molten fluid. Former two are mostly same (only difference i know is when in wooden pipes) and have no "molten." prefix. So your code would not work for helium plasma.
- On the other hand, molten sulfur does not exist. This is not a bug, but why don't they produce sulfur dust (or something similar) instead?
- Conversion ratio of Air-to-Molten fluid should not be 1L plasma for 1L molten, but 125L for 16L.
- No documents, no bugreports. We (at least I) cannot see what you actually want via the code, even if you are super-programmer. And please do test and do documentation before PR if you are good programmer. ;-)
- Discussion of plasma turbine output is #162 .

Sorry for my poor engrish. Hopefully someone follows.

@Blood-Asp please rollback it.

Blood-Asp:
The oxygen, helium and nitrogen do work. Molten is only tried when these do not work.
There is no fluid version of sulfur and no way to handle non fluids for the GT5 fusion reactor, so that will stay at no output.
The conversion ratio for plasma -> molten is something i plan to add, but for now it is not that bad, since the outputs are all cheap materials and this is endgame tech. No balance broken here.
The discussion about this PR was on the forum. This code was running on a server for months now and looking at it it seems to be fine.

ShelfOwl:
Okey, I apologize you to doubting not to test and consideration.

PR #567, opened 2016-07-15
Added Superconducting Coil as coil for Multi-Smelter and Blast Furnace (for end-game recipes)

OneEyeMaker:
Changed energy requirements and work speed for Multi-Smelter
Set values for Superconducting coil twice more, than for regular coils:
- Heat Capacity = (Nichrome coils heat capacity + 900) \* 2;
- Smelting speed = (Nichrome coild smelting speed \* 2) \* 2.


Blood-Asp:
Superconductor heating Coils does not make sense. They can not handle heat and have to be cooled. They are meant to create strong magnetic fields.
When you want to add Coils for higher temperatures you should add new Tungsten Coils or something like that.

PR #586, opened 2016-07-29
Merge GalacticGreg into GT

### LINKED ISSUE #494: Suggestion: Make GC asteroid blocks yield something valuable
Its strange that moon dust and mars stone yield usefull stuff, while the asteroids are only decorative. Some have an ore core, but thats not realy usefull. They could yield rare metalls like platinum or larger amounts of iron or carbon...

mura3277:
@Blood-Asp, Let me know if I've missed anything.

This should remove the requirement of GalacticGreg. Ore generated in Galacticraft dimensions with GalacticGreg should not be effected.

-Added processing recipes for Asteroid blocks, Fixes #494.

If you're playing with GT6, you will still need to use GalacticGreg.


mura3277:
So, for some reason, the dimension section of the WorldGeneration config is not getting set properly for Moon, Mars and Asteroids. @Blood-Asp, do you know what I could of broken?

Edit: The main section for mixes/smallore seems to be correct(GC dims are correctly true or false), but the dimension section is not. I guess I'm not understanding how that section of the config works?

src/main/java/gregtech/common/GT_Worldgen_GT_Ore_SmallPieces.java
```diff
@@ -42,7 +42,7 @@ public boolean executeWorldgen(World aWorld, Random aRandom, String aBiome, int
         if (!this.mRestrictBiome.equals("None") && !(this.mRestrictBiome.equals(aBiome))) {
             return false; //Not the correct biome for small ore
         }
-        if (!isGenerationAllowed(aWorld, aDimensionType, ((aDimensionType == -1) && (this.mNether)) || ((aDimensionType == 0) && (this.mOverworld)) || ((aDimensionType == 1) && (this.mEnd)) || ((aDimensionType == -28) && (this.mMoon)) || ((aDimensionType == -29) && (this.mMars)) || ((aDimensionType == -30) && (this.mAsteroid)) ? aDimensionType : aDimensionType ^ 0xFFFFFFFF)) {
+        if (!isGenerationAllowed(aWorld, aDimensionType, ((aDimensionType == -1) && (this.mNether)) || ((aDimensionType == 0) && (this.mOverworld)) || ((aDimensionType == 1) && (this.mEnd)) || ((aDimensionType == -28) && (this.mMoon)) || ((aDimensionType == -29) && (this.mMars)) ? aDimensionType : aDimensionType ^ 0xFFFFFFFF)) {
```
OneEyeMaker: // I guess, you shouldn't use hardcoded -28, -29,... dimension IDs. These IDs are configurable in GC.

mura3277: Ah you're right, I didn't know it was configurable, thanks for pointing that out.

PR #594, opened 2016-08-07
Paintable Wires

Technus:
Fully tested, solution to make wires (not cables) visibly paintable (no more guessing the color).


Dream-Master:
@Technus
Is this the code we have allready in GT New Horizons ?

Technus:
Yes. that is why i said it is tested

Dream-Master:
K

PR #620, opened 2016-09-03
Fixing efficiency calculation operator precedence.

Technus:
This will allow more eu to b eproduced by simple matter of:

EUt_Eff/10000 -> EUt_(Eff/10000)

Saving a lot of Integer space.


Blood-Asp:
What exactly are you trying here? All i see is that you make the math less accurate by dividing the integer first and cutting off decimals.

Technus:
If u produce EU at rates near to Integer_max multiplying it first makes the integer overflow to negative values

Technus:
Afaik it only occurs in large gens

Blood-Asp:
then make it long instead of integer. but the way you change it here breaks calculation for small values.

Technus:
can do

Technus:
Done

Bogdan-G:
idea: `Eff/10000` - hmm it is possible to replace the bit operation? (:P) I as that didnt notice these things especially

PR #632, opened 2016-09-13
Various Bugfixes

mura3277:
Credit goes to @Bogdan-G


Bogdan-G:
reappeared in notifications, what is happening here? I do not see any changes

Issue #718, opened 2016-11-01
Vein centers in negative coordinates are 1 chunk closer to axes than they should be


MauveCloud:
This issue actually predates the unofficial fork, and occurs in 5.08, 5.09, and 5.10, but Greg fixed it for GT6 back in May 2015 (ref: http://forum.industrial-craft.net/index.php?page=Thread&postID=187659#post187659)
Current code is this:
```java
if ((Math.abs(this.mX / 16) % 3 == 1) && (Math.abs(this.mZ / 16) % 3 == 1))
```

However, that results in a pattern like this:
```
Math.abs(-160 / 16) % 3 = 1
Math.abs(-144 / 16) % 3 = 0
Math.abs(-128 / 16) % 3 = 2
Math.abs(-112 / 16) % 3 = 1
Math.abs(-96 / 16) % 3 = 0
Math.abs(-80 / 16) % 3 = 2
Math.abs(-64 / 16) % 3 = 1
Math.abs(-48 / 16) % 3 = 0
Math.abs(-32 / 16) % 3 = 2
Math.abs(-16 / 16) % 3 = 1
Math.abs(0 / 16) % 3 = 0
Math.abs(16 / 16) % 3 = 1
Math.abs(32 / 16) % 3 = 2
Math.abs(48 / 16) % 3 = 0
Math.abs(64 / 16) % 3 = 1
Math.abs(80 / 16) % 3 = 2
Math.abs(96 / 16) % 3 = 0
Math.abs(112 / 16) % 3 = 1
Math.abs(128 / 16) % 3 = 2
Math.abs(144 / 16) % 3 = 0
```

Which produces excess overlap near the axes.  However, if one adds a large multiple of 3 after dividing by 16 instead of taking the absolute value, proper spacing is maintained:
```
(-160 / 16 + 402653184) % 3 = 2
(-144 / 16 + 402653184) % 3 = 0
(-128 / 16 + 402653184) % 3 = 1
(-112 / 16 + 402653184) % 3 = 2
(-96 / 16 + 402653184) % 3 = 0
(-80 / 16 + 402653184) % 3 = 1
(-64 / 16 + 402653184) % 3 = 2
(-48 / 16 + 402653184) % 3 = 0
(-32 / 16 + 402653184) % 3 = 1
(-16 / 16 + 402653184) % 3 = 2
(0 / 16 + 402653184) % 3 = 0
(16 / 16 + 402653184) % 3 = 1
[...]

Issue #705, opened 2016-10-25
Incorrect Maceration/Pulverization recipes


Timeslice42:
(vanilla) Stone furnaces **Macerate** into 8 stone dust (which is correct), however they **Pulverize** into 8 sand (which is wrong). According to IC2 the only stone related item which is supposed to grind into sand is cobblestone; everything else (stone, slabs, machines, etc.) grinds into stone dust.

Nex4rius:
You could always turn furnaces into sand in a macerator.

Blood-Asp:
I somehow want to keep the furnace = 6 sand recipe... always used it for mass sand production.

PaladinOne:
Maybe some sort of recipe that converts Stone Dust into Sand?

Timeslice42:
@Nex4rius I'm actually trying to turn stone furnaces into stone dust, but sadly the steam macerator is a pulverizer, not a macerator. Or something...

yznpku:
I believe this recipe is actually a feature.

Bogdan-G:
ancient bug, i have seen it 2 years ago.
2 recipe conflict and overlap randomly at each other at every startup. The solution is simple: to regulate all the recipes, or check and remove for solve the conflict. Not only ic2 adds cobblestone crushing recipe in the sand in crushers.

ps in IRL sand and stone dont 1 type.

Timeslice42:
I'm not sure what "dont 1 type" means, but in the same way that gravel isn't sand, sand isn't dust. Think of stone dust as crushed sand.

But yeah, it's the fact that the recipes are inconsistent that bothers me the most.

MauveCloud:
Personally, I think something like this would make somewhat more sense:
Anything made with a significant amount of stone/cobble (including furnaces and straight cobblestone) in a macerator/pulverizer produces stone dust.
Cobblestone can already be hammered for gravel.
Gravel can be forge hammered to get sand, or sifted to get flint (and maybe also macerated/pulverized for stone dust), instead of macerated for flint and washed for stone dust. (I mean, how does that make any sense?  water reduces it to dust, while a macerator with a diamond grinding head leaves a big piece of flint behind?)

Bogdan-G:
@Timeslice42 

> "dont 1 type"

[...]

PR #755, opened 2016-11-14
Fix for negative-coordinate vein generation, adjustments to recipes for stone dust, sand, and flint

MauveCloud:
Addresses #718 and #705

Vein centers will now be at roughly ... -120, -72, -24, 24, 72, 120, ... on the x and z coordinates instead of ... -104, -56, -8, 24, 72, 120, ...

Pulverizing furnaces or cobblestone will now produce stone dust (for consistency, similar to cobblestone slabs, or droppers/dispensers made mostly from cobblestone), and so will pulverizing gravel.  However, gravel can be sifted (with a scoop in early game, or with the electric machine later) to get flint, or hammered to get sand.


codewarrior0:
> hammered to get sand.

Do you happen to know if this will allow the JackHammer (HV) to break Gravel now? Or is it just a crafting recipe with the Hammer?

Dragon2488:
Blood has already said that furnace to sand recipe is correct and shouldnt be changed.

MauveCloud:
> Do you happen to know if this will allow the JackHammer (HV) to break Gravel now? Or is it just a crafting recipe with the Hammer?

I haven't checked that yet.  Having the Forge Hammer recipe allows a regular hammer tool to break gravel into sand (even if such a recipe was added by Minetweaker), dunno about the JackHammer.

Edit: the JackHammer does not break gravel.  If you want me to add that, please provide justification.

> Blood has already said that furnace to sand recipe is correct and shouldnt be changed. 

Not exactly.  What he said in the bug was "I somehow want to keep the furnace = 6 sand recipe... always used it for mass sand production."  That isn't a claim that the recipe is correct, or a justification for the inconsistency, and the Forge Hammer recipe provides an alternative for mass sand production from cobblestone.

Timeslice42 (commented):
You have some whitespace inconsistancies in several places.

MauveCloud:
> You have some whitespace inconsistancies in several places.

I think that's an artifact of my editor settings - I hadn't noticed that it was adding tabs instead of spaces, which GitHub expands to width 8 instead of 4.  I'll see if I can fix that.

Blood-Asp:
Now conflics with previous PR...

PR #774, opened 2016-11-26
Adjust Diesel Engine tooltip to match requirements

codewarrior0:
16 Casings are required along the corners and edges. The Hatches may not be along a corner or edge, but may only be placed on the sides, next to a Gear Box.

Also note Diesel, Lube, and Oxygen use.


codewarrior0:
BTW, there's no limit on the maintenance or input hatches, so you could save 32 Titanium by replacing the four remaining unused casings on the sides with hatches. Not quite as big a deal as the 14 extra hatches on the distillery.

PR #777, opened 2016-11-28
Fixed #767

### LINKED ISSUE #767: Gas Turbine with infinite fuel
If put empty Universal Fluid Cell to "In" slot of Gas Turbine with some amount of any fuel it moves to "Out" slot and stays empty. But amount of fuel in turbine starts to increase infinitely.
gas turbine bug

Dimach:
Industrialcraft universal fluid cell produce so much bugs. Please not use it in recipes.


Blood-Asp:
This "fix" does only remove the broken universal cells without any replacement. That way it is only half done.

Dimach:
Nope, because i fixed incorrect behavior of GT_MetaTileEntity_BasicGenerator. If now return universal cell in recipes fuel in generator will not grow to infinity, its first. Second - recipes with universal cell not worked at all, any UFC (even empty) generator will receive it like full of rocket fuel and try to burn it. I thing its problem of industrial craft, not gregtech. And third - recipe with rocket fuel already added by common cell and no need to use universal.

Blood-Asp:
At last the Magneticraft Natural Gas is still missing...

PR #778, opened 2016-11-28
Ore harvest level rework. Fixed #730.

### LINKED ISSUE #730: Flint Pickaxe vs Small Ores
Small Tine Ore in Blue Shcist stone can't be mined neither by Flint nor by even Iron pickaxe.

Dimach:
Additional parameter in getHarvestData needed because UndergroundBiomes metadata was intersected with gregtech granite and cause bugs like #730. Now possible to add more ore blocks with no metadata intersections.


Dimach:
And it not fix already generated block, only generated after installing fixed version.

Issue #735, opened 2016-11-06
Adv Miner II dupes Oilsand.


SapientMC:
While mining Oilsand Adv Miner II add 1-5 blocks of oilsand per operation to output bus. Bus it does not take away blocks from Overworld. Mining pipe even does not move down after more than 1000 Oilsand blocks mined.

codewarrior0:
I can't confirm that it happens now, but this did not happen in 5.09.21.

SapientMC:
i'm on latest 27pre

codewarrior0:
Dunno if it's relevant, but I just let an advminer run for a couple of hours and found two oilsand blocks in its output. There were no oilsand veins within 100 blocks of the thing.

SapientMC:
@codewarrior0 may be you mined it somewhere else during AdvMiner working?

codewarrior0:
Yes, I did! And I found out exactly what is happening. Patch incoming.

Issue #674, opened 2016-10-08
Advanced Miner II bug


Antifluxfield:
5.09.22&.23, multiple adv miners "share" the ores. The problem is similar to, or maybe the same as #508.
Moreover, when mining with laser, drill or pickaxe, with the adv miner working at the same time, it seems that the adv miners also output the crushed version of the manually excavated ores. For example, my adv miner II was working on a naquadah vein in the end, and I knocked off some phosphorus in overworld. Later I found some crushed phosphorus and a apitite crystal in the miner's output, which should never occur in the end.

Blood-Asp:
All i can say about that is: I have no idea. Tried it a few times already and never found anything.

Issue #508, opened 2016-05-04
Advanced Miner II is magic


zaitsevyan:
It always mines something interesting: NetherQuartz in overworld, crushed diamond ore at 40 lvl pipe, Ores that are not in current location of miner(huge amount of it, it is not small ores.), etc. I am using Infitech 2 modpack.

Blood-Asp:
What seems to happen is, that multiple miners in one world seem to link up and share found materials...

DerEchteLanger:
Thats not the case, on the official server we are the only one with such a miner and still got even naquadah in the overworld...

zaitsevyan:
@Blood-Asp There was only one miner in the world.

PaulMadsen:
I can confirm the ores are shared.  I am not sure, but it appears that they will truncate ores when they share it.  As if one will overwrite what the other is pulling up.  I have not tested this.  It just appears to be this way.

SapientMC:
Seems like its same bug as https://github.com/Blood-Asp/GT5-Unofficial/issues/469

Blood-Asp:
https://github.com/Blood-Asp/GT5-Unofficial/pull/533 maybe fixed?

PR #782, opened 2016-12-05
Fix Advanced Miner II occasionally outputting the wrong blocks

### LINKED ISSUE #751: Multiple advanced miners mixed output
I build two advanced miner setups. One on tungsten ore, and other on platinum ore. Ore drops in little ae setups, and i see, that i get sheldonite ore on tungsten deposit, and scheelite on platinum deposite. Both ore mixes are rare, and i check around chunks for ores. No platinum near tungsten, and no tungsten near platinum.

codewarrior0:
AdvMinerII would occasionally output the drops from a block recently
broken by the player. This happens when the wrong coordinates are
passed to GT_Block_Ores_Abstract.getDrops(), which causes it to fail to
find the TileEntity and then fall back to whatever the most recently
broken block was.

Fixes #751 (probably), #735 (probably), #674, #508, maybe more!

I have a feeling that the only way AdvMinerII was ever outputting the correct drops was by going through the Silk Touch codepath.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_AdvMiner2.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_AdvMiner2.java
@@ -134,7 +134,8 @@ public class GT_MetaTileEntity_AdvMiner2 extends GT_MetaTileEntity_MultiBlockBas
             }
             ArrayList<ItemStack> tDrops = new ArrayList();
             Block tMineBlock = null;
-            ChunkPosition mle = null;;
+            ChunkPosition mle = null;
+            int posX, posY, posZ, offX, offY, offZ;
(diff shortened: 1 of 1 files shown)


Techlone:
I fixed it few months ago. Probably my changes has been overwritten. I don't know how it's happened.
#533

codewarrior0:
(in response to a deleted comment referencing #533)

Your changes were overwritten when the Miner was changed to be able to mine any kind of ore, not just GregTech ores. In your patch, you use a GregTech-specific function to get the drops for the ore, but in the latest revision, the standard Minecraft function is used instead, and is passed the wrong coordinates.

Techlone:
@codewarrior0 I just looked my code and it isn't correct. At that date I understood the problem but wrote wrong code...
Or not I'm confused

Techlone:
Nevermind. Present PR resolve the problem. Only this is important.

PR #799, opened 2016-12-13
Add progress bar

danbka33:
Add progress bar
eao6vglsx47wz2


Blood-Asp:
@danbka33 What is the Forge version requirement for this? I did try the same quite some time ago and back then the forge implementation was incomplete. Pretty sure it needs a quite up to date build. Might be nessary to make this optional depending on forge version or put this into a try catch.

@Dragon2488 I would prefer to merge PRs myself as i'm the one to provide new builds anyways.

danbka33:
@Blood-Asp Forge 10.13.3.1399+

PR #802, opened 2016-12-14
Don't allow Large Turbine to start up without Dynamo Hatch

codewarrior0:
Large Turbine was incomplete if the Dynamo Hatch spot was taken up by a different GTTE, but not if that spot was just Air or some other block.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_LargeTurbine.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_LargeTurbine.java
@@ -94,6 +94,8 @@ public abstract class GT_MetaTileEntity_LargeTurbine extends GT_MetaTileEntity_M
                 } else {
                     return false;
                 }
+            } else {
+                return false;
             }
         } else {
             return false;


SapientMC:
2kV Dynamo hatch on Large Diesel Generator in Turbo mode does not Explode.
Could you please fix it too?

codewarrior0:
Hum, I'll bet it's comparing the mEUt pre-mEfficiency-multiplier against the Dynamo Hatch's tier voltage...

codewarrior0:
It doesn't look like the Diesel Engine is meant to explode. I'll have to defer to @Blood-Asp on that one.

Actually, I should ask @Muramasa- since he contributed the engine.

SapientMC:
All other multiblock turbines explodes when produce more energy than Dynamo Hatch can transfer.
Think Large Diesel Engine should work same.

codewarrior0:
On the other hand, the Diesel Engine is the "odd one out" - it doesn't use Turbine Rotors, consumes heavier kinds of fuel, and has only two fixed power levels rather than the extreme ranges of power given by the various Rotors.

mura3277:
The dynamo should probably explode. And yeah the two fixed power outputs is odd. I originally wanted to allow players to change the RPM of the engine so they can balance fuel consumption and power output while having different "sweet-spots" for the fuels.

Issue #769, opened 2016-11-22
Crash with Modular Exoskeleton


JasonMcRay:
#### Issue Description:
I'm using SEUS shader.

I was trying to use Modular Exoskeleton with some items (described below).
It eventually crashed

#### Is the issue happening on server? Client? Both?
When I get this crash on multiplayer server, I see 'Internal Server Error'  and on single player, it just crashed. Attached log was generated at single player

#### What happens:
CRASH!

#### What you expected to happen:
Just wanna use them :)

#### Steps to reproduce (if aplicable):
I wear 4 modular exoskeleton with
Medium Lithium Battery
MV Electric Motor
Good Elecronics
Aluminium plate

____
#### Your most recent log file where the issue was present: 
http://pastebin.com/yEr0QN6i

PR #803, opened 2016-12-14
Crash with Modular Exoskeleton #769

dgabehar:
Fix for Crash with Modular Exoskeleton #769. Removed player.closeScreen() call in GuiElectricArmor1 and GuiModularArmor to eliminate the graphics glitch that happens when switching between Modular Armor inventory tabs.
Fixed issue with Modular Armor inventory not saving properly after removing the call to closeScreen
Added null and key checks when determining if a full radiation or electrical armor set is configured


Blood-Asp:
Still working on this or ready to merge?

dgabehar:
Done with it for now. Won't have time to work on it for a couple of weeks.

PR #858, opened 2017-01-13
Disable miner when mining pipes is over

Techlone:


codewarrior0:
I *thought* it was weird that `moveOneDown` returned something that never got checked.

Techlone:
I didn't see the drilling rig. Maybe it has a similar bug.

PR #866, opened 2017-01-22
1.10.2 misc fixes

codewarrior0:
Fixes Charcoal Pile Igniter, incorrect colors on metatools, a crash when rendering a newly crafted Soldering Iron (and possibly every empty electric tool), and stops filling up the console with blank lines.


codewarrior0:
I'm also holding on to some much less conservative changes, related to connecting fluid pipes to the new IFluidHandlers, rendering GT blocks in a thread-safe way (to fix several rendering bugs), and removing a lot of unneeded chunk updates. I'm still testing them on a private server to see how they play with a few other mods we're using, including TConstruct and Integrated Tunnels.

PR #893, opened 2017-02-10
Reactor chambers support

Techlone:


Nex4rius:
Does this mean we can put cables directly on the reactor chamber without ic2 blocks inbetween?

draknyte1:
You can only attach a transformer, as the reactor still outputs IC2 EU.

PR #904, opened 2017-02-17
Add new semiconductors and indium recipe

Techlone:
This new materials needed for our vision of solar panels recipes. Maybe it will be useful for your recipes too when (or if) you will decide to remake solar panels.
PR is including indium processing and two new materials - Gallium arsenide and Indium gallium phosphide


Blood-Asp:
Will look again at this once i have done my own circuit changes. Mostly only missing textures for that so almost done.

PR #947, opened 2017-03-07
Resolve #917.

### LINKED ISSUE #917: Reset cleanness of Clean Room
If door are opened cleanness should reset to zero and stay so until its closed. After should slowly rise to 100%.

Dimach:
Cleanroom now reset cleanness if door open.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_Cleanroom.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_Cleanroom.java
@@ -134,7 +134,7 @@ public boolean checkMachine(IGregTechTileEntity aBaseMetaTileEntity, ItemStack a
 	}
 
 	if(doorState){
-		mEfficiency=0;
+		mEfficiency=Math.max(0,mEfficiency-200);
 	}
 
 return true;
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_Cleanroom.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_Cleanroom.java
@@ -56,6 +56,8 @@ public boolean checkMachine(IGregTechTileEntity aBaseMetaTileEntity, ItemStack a
 	int mDoorCount=0;
 	int mHullCount=0;
 	int mPlascreteCount=0;
+	boolean doorState=false;
+	mUpdate = 100;
 	for(int i = 1;i<8;i++){
 		Block tBlock = aBaseMetaTileEntity.getBlockOffset(i, 0, 0);
 		int tMeta = aBaseMetaTileEntity.getMetaIDOffset(i, 0, 0);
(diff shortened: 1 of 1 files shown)


MauveCloud:
Does this consider the possibility of there being an airlock set up, or require protective clothing?  Also, I'm not so sure it should instantly reset the efficiency to zero - seems more like it should decrease based on how large the cleanroom was built relative to the door size and how long the door stays open.

Dimach:
Airlock setup not allowed. Protective clothing not needed because multiblock check only door state, not player inside.
About instantly reset - maybe better decrease cleanness by one percent every 5 second while door stay open?

Blood-Asp:
1% per 5 sec sounds fine. 
Airlock is a bit complex to code but would be nice. Check for player with protective clothing should be easy for me to add later myself too.

Dimach:
Done.

SapientMC:
@Blood-Asp >1% per 5 sec sounds fine.
Such speed could be used in case lost of power supply for clean room. But opened door should reset cleanness completely.

That makes necessity to full automated production inside clean room and prevent manual crafting.

leagris:
##### About [automation requirement](https://github.com/Blood-Asp/GT5-Unofficial/pull/947#issuecomment-284985789) from @SapientMC

> That makes necessity to full automated production inside clean room and prevent manual crafting.

Automation means are very limited inside clean-room; unless it has means to transport items, fluids and, control precisely the amount of material inserted and extracted from machine inside the room. It needs to pass these across room walls.

Means having options settings for custom wall blocks whitelist to enable various pipes, conduits AE2 cables, including multi-part stuffs (except those MCMultipart parallel covers that can let player (and unwanted dust) to trespass.

Also the Clean-Room must emit a Redstone signal, with strength matching cleanness level, so to automate start of production at optimal level.

A cosmetic indicator light could be rendered as an overlay over the controller as a cosmetic visual of the cleanness state.

leagris:
A blocks white-list could also allow other kind of doors, not only vanilla/IC2.
Malisis Doors can provide very nice laboratory themed doors.

Alternatively, Gregtech could provide its very own nice looking airlock, window and lighting features to integrate into walls, ceiling, floor. (S.O. these Redstone controlled light blocks from Gregtech 4 and earlier)

SapientMC:
> It needs to pass these across room walls.

There are no problem with that. Machine Hull + 2 conveyor module can transfer items through the wall. Hull + 2 pump covers can transfer fluids.
Now there are only one machine need to be placed inside clean room - Laser Engraver. And it has only one input and one output slots for items.

leagris:
How do you pass Redstone signal to and from the Laser Engraver in the room from/to outside the Room?

SapientMC:
What for do you need it?

PS Wireless redstone transmitter/receiver

leagris:
Control machine covers. Machine work cover, control cover…

SapientMC:
Why do you need them for laser engraver?

leagris:
And if Clean-Room has only Laser-Engraver need. Why even create a room and not a multi-block. What is the point of a multi-block single-purpose machine you may enter inside if you are not supposed to enter inside for anything?

leagris:
Why are there these covers if I don't need these?

SapientMC:
Covers are needed to automate large boilers for example. And another multi-blocks, but not single block machines. You can just regulate input.

DEMOOH:
It is possible to automate cleanroom. Machine hulls can accept energy and items, so there is no restriction for that. And you can use ae2 rings.

leagris:
Covers are also used to monitor machine work, monitor machine power, control machine stop/start...

leagris:
Monitoring machine work is a perfectly valid scenario so you don't open the room door while it is processing a batch of products.
Control machine stop/start is perfectly valid to stop production before entering the room and until the room is clean again.

DEMOOH:
I am pretty sure that you can send redstone signal through the wall. The old vanilla block-torch trick or something more appliable.

StragaSevera:
If we will have full cleanness reset, we will need some device to monitor state of the cleanroom.
By the way, I think, cleanness should decrease if there is a person inside without some protective clothes...

DEMOOH:
I think that this discussion should move here https://github.com/Blood-Asp/GT5-Unofficial/issues/917

SapientMC:
> If we will have full cleanness reset, we will need some device to monitor state of the cleanroom.

NC Info Panel can show it.

leagris:
GT Casings don't transmit Redstone.

SapientMC:
> GT Casings don't transmit Redstone.

GT has wireless redstone transmitter/receiver covers.

Blood-Asp:
the GT scanner also shows cleanness (efficiency).
Transfer inside/out works just fine. Did anyone even try before complaining?

leagris:
> GT has wireless redstone transmitter/receiver covers.

Wireless Redstone transmitters are a poorly designed work-around Gregtech5 feature regression.

> the GT scanner also shows cleanness (efficiency).

The GT scanner a debug tool. It is unable to Redstone signal machine states or control machine operations.

@Blood-Asp I like new features you add.
What I don't like is the more grind and stuffs that can not be automated and monitored setup and forget.

This is the reason. If a bug-fix-only release was available I'd update Gregtech for the Beyond Reality pack. Otherwise, it is again more grind for new players, and trouble for players on existing worlds.

MauveCloud:
> I am pretty sure that you can send redstone signal through the wall. The old vanilla block-torch trick or something more appliable.

I presume you're referring to the redstone torch, but this reminds me - regular torches should probably corrupt the cleanness of the cleanroom, encouraging players to use glowstone, luminators, or light sources from other mods (e.g. flood lights).

StragaSevera:
I tried it - plascrete does not permit this trick. The only way is wireless redstone.

Blood-Asp:
Somehow no GT blocks allow that. But no idea why. If someone could find the reason...

OvermindDL1:
> Somehow no GT blocks allow that. But no idea why. If someone could find the reason...

Is `isProvidingWeakPower` not set properly?

Are they testing that they are running 'strong' redstone to the block and detecting if 'weak' redstone is leaking through?

Is the block not marked as transparent or non-opaque or non-fullblock?

Few other things to check.  ^.^

PR #952, opened 2017-03-11
Config for using only soldering alloy

Techlone:
This changed check gets a config value by name `recipes.harderrecipes.useonlygoodsolderingmaterials` and doesn't add recipes with tin and lead if this cfg value true. I think it's useful. Default it's `false` and doesn't change recipes.


Techlone:
Don't pay attentions on merge commits

Techlone:
Hm, recipes of adv circuit still use tin and lead. Don't merge the PR now please.

Techlone:
Done

leagris:
An other option is to add lead-free soldering alloy that fail after some time :rofl:

PR #960, opened 2017-03-14
Kill excess redraws

codewarrior0:
Eliminates pointless redraws when server state is synced to the client, by checking whether the state was actually changed before issuing the redraw.

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/BaseMetaPipeEntity.java
+++ b/src/main/java/gregtech/api/metatileentity/BaseMetaPipeEntity.java
@@ -350,6 +350,7 @@ public class BaseMetaPipeEntity extends BaseTileEntity implements IGregTechTileE
 
     @Override
     public boolean receiveClientEvent(int aEventID, int aValue) {
+        boolean tNeedsRedraw = false;
         super.receiveClientEvent(aEventID, aValue);
 
         if (hasValidMetaTileEntity()) {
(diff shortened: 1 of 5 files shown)


src/main/java/gregtech/api/metatileentity/BaseMetaPipeEntity.java
```diff
@@ -362,10 +363,13 @@ public boolean receiveClientEvent(int aEventID, int aValue) {
...
             switch (aEventID) {
                 case 0:
-                    mConnections = (byte) aValue;
+                    if (mConnections != (byte) aValue) {
+                        mConnections = (byte) aValue;
+                        tNeedsRedraw = true;
+                    }
                     break;
                 case 1:
                     if (hasValidMetaTileEntity()) mMetaTileEntity.onValueUpdate((byte) aValue);
```
codewarrior0: wtf

MauveCloud:
I tried building a local copy of this, manually merged with my pending PR, but discovered that while it prevents machines from flickering at a distance, glitched textures become static, similar to GT ores and GT stone, so that even close up, machines and cables could have corrupted textures that stayed that way, instead of stabilizing as I got closer.

codewarrior0:
Yeah, it doesn't fix the glitched textures - in fact, it makes it harder to isolate that bug. The high redraw rate is what helps to expose it. Like I mentioned in the other issue, the glitched textures are related to threaded rendering and having state stored in a singleton object that's shared among the rendering threads.

MauveCloud:
Does that mean you have a lead on a better fix for the glitched textures?  If so, why is this PR still open?

codewarrior0:
Because it fixes a separate issue: excessive redraws caused by machines, wires, and pipes.

MauveCloud:
😕 I guess it's ultimately up to Dragon2488 or Blood-Asp whether to accept this PR, but my opinion is: until the glitched textures issue is properly fixed, this change makes things look worse.

Blood-Asp:
Maybe add some reduced default redraw? Like once every 100 ticks? Should fix glitched textures after a few secs but still be less laggy.

codewarrior0:
No.

Dragon2488:
These random updates were my attempt to force server-client tile entity sync or early stage of port. I don't think it's really needed now, but for temporary flickering fix it should stay.

PR #980, opened 2017-03-22
BoP Nether fix

MauveCloud:


Blood-Asp:
Seeing how much fixes you commit, want to be added as Collaborator?

Blood-Asp:
But best for this commit would be to keep old behavior. aka: Only End or Nether to as dimension 1 or -1, every other as 0 spawning same as overworld.

MauveCloud:
> But best for this commit would be to keep old behavior. aka: Only End or Nether to as dimension 1 or -1, every other as 0 spawning same as overworld.

Except that's not really what the old behavior was doing.  Only _vanilla_ End or Nether was being treated as dimension 1 or -1 respectively.  BoP End and Nether were inheriting overworld spawn type, and so were Galacticraft planets and asteroids, despite the WorldGeneration file having separate entries for them in each small ore type and mix vein.

Blood-Asp:
Well, to rephase it, every dimension not handled differently should behave as 0/overworld. Other dimensions should, if possible, have by default GT overworld oregen. For example Twilight Forest or Deep Dark. Suddenly not spawning GT ores there will annoy most people more than not having correct ores in BoP Nether.

MauveCloud:
Okay.  How's this?

Blood-Asp:
That should be fine.

PR #981, opened 2017-03-24
Fixed various Bugs on the Powderbarrel.

dohyun22:
Fixed PrimedTNT Entity being spawned on client side when destroyed by player.
Fixed PrimedTNT Entity spawning at misplaced location (about 0.5m from the center of the block).
Fixed fuse sound not being played when ignited.
Fixed Powderbarrel not being able to spawn PrimedTNT Entity when destroyed by explosion.

### DIFF
--- a/src/main/java/gregtech/common/blocks/GT_Block_Reinforced.java
+++ b/src/main/java/gregtech/common/blocks/GT_Block_Reinforced.java
@@ -216,7 +216,7 @@ public class GT_Block_Reinforced extends GT_Generic_Block {
     public boolean removedByPlayer(World world, EntityPlayer player, int x, int y, int z)
     {
       if(!world.isRemote && world.getBlockMetadata(x, y, z)==5){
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/blocks/GT_Block_Reinforced.java
```diff
@@ -215,10 +215,10 @@ public void dropBlockAsItemWithChance(World aWorld, int aX, int aY, int aZ, int
     
     public boolean removedByPlayer(World world, EntityPlayer player, int x, int y, int z)
     {
-      if(world.getBlockMetadata(x, y, z)==5){
-    	EntityTNTPrimed entitytntprimed = new EntityTNTPrimed(world, x, y, z, player == null ? null : player);
```
OneEyeMaker, line 219: ```player == null ? null : player``` always return ```player``` (in any case). You can simplify this code.

PR #993, opened 2017-03-30
Cleanup

Exidex:
Code cleanup


draknyte1:
Some of the code changes are dumb. On what basis is it a cleanup?
Removing privates and ; from enums gains no performance and doesn't help readability...

Exidex:
This changes were made due to uselessness of this privates and ;

draknyte1:
Try fix some proper issues instead of making the code harder to read as This 'useless' code adds to readability. It offers no benefits to performance at all and I feel as though you've just used something like eclipse's 'code cleanup' feature.. 

Any valid reasons why a merge should be considered? private modifiers arent useless when you don't want other mods to access functions...

codewarrior0:
Changing access level is not cleanup. 

If this were my project, I wouldn't merge this in a million years.

Exidex:
This is private access on enum constructors which are implicitly private, but I get your point

PR #998, opened 2017-03-31
Remove PFAAGeologica mod support.

Exidex:
Remove PFAA Geologica mod support due to mod not being updated to 1.10.2.


MauveCloud:
The PFAA Geologica author has not given up on the mod.  According to this post:
http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/1293751-per-fabrica-ad-astra-geologica?comment=629
There are plans for a 1.10.2 version of PFAA Geologica, so it might be premature to remove PFAA support from GT5u.

Blood-Asp:
Exactly. There is no real gain in this PR. Maybe a bit code cleanup, but there are much better points for that.

PR #1001, opened 2017-04-01
Fixed right click behavior.

Exidex:
Fixed right click behavior for metaitems.


src/main/java/gregtech/api/items/GT_MetaBase_Item.java
```diff
@@ -143,16 +145,19 @@ public EnumActionResult onItemUse(ItemStack aStack, EntityPlayer aPlayer, World
...
+                    if (aStack.stackSize <= 0) aPlayer.setHeldItem(hand, null);
+                    return EnumActionResult.SUCCESS;
+                }
             }
+        }
         if (aStack.stackSize <= 0) {
             aPlayer.setHeldItem(hand, null);
             return EnumActionResult.SUCCESS;
         }
-        return EnumActionResult.SUCCESS;
```
codewarrior0, line 155: This change looks like it will fix https://github.com/jaquadro/StorageDrawers/issues/369

PR #1003, opened 2017-04-02
Fixed #1000.

### LINKED ISSUE #1000: [5.09.29] Crash while MC loading
gregtech-5.09.29pre2.jar works fine
gregtech-5.09.29.jar - crashes
https://pastebin.com/j0L5iWzP

Dimach:
Yep, i used deprecated api, but it supported in java7.

### DIFF
--- a/src/main/java/gregtech/GT_Mod.java
+++ b/src/main/java/gregtech/GT_Mod.java
@@ -54,8 +54,6 @@ import net.minecraftforge.oredict.OreDictionary;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.*;
-import java.time.LocalDate;
-import java.time.ZoneId;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


draknyte1:
Please ensure it works with latest J8. 8 support is far more important than 7, why people would even still use 7 is beyond me..

draknyte1:
April Fools is also April 1st, not month three. Unless the older API has January at 0..

Blood-Asp:
Yes, the older one has january at 0. Tried that one before myself but exactly because of that depricated i changed that. 
But adding a try/catch because of that depricated would maybe be a good idea. since it is posible to be removed in some future java version and whould then crash again.

Dragon2488:
Fuck, please don't do shit. It can be easily replaced with new Calendar API.

Dragon2488:
Fixed it myself, maybe getBoolean shouldn't be there. But fuck, code properly or don't code at all.

draknyte1:
Supporting J7 imho is stupid when it's so outdated. The fix is update Java, not support old Java.

Bogdan-G:
^ J8 no need, no perfomance, new bugs and etc :D 
PS goto J9 xD

MauveCloud:
> PS goto J9 xD

That's difficult to agree with when Java 9 is still in "Early Access" stage (and from what I can find, the latest estimate for an official release is July 2017)

draknyte1:
J7, still has outstanding issues that will never be fixed. API components 8 replaced are faster, including the entire JS engine. Supporting deprecated API doesn't really benefit anyone when J8 is the standard, even for EE.

Bogdan-G:
>EE

J2EE or mod EE?
PS (for everyone) about the pull-request: why create a object calendar? There is an variant without dependency API.

PR #1007, opened 2017-04-05
Fixed IntelliJ thinking it was Java 6

ghost:
I added something to the gradle file so that IntelliJ IDEA reads the project as java 8 instead of java 6.


draknyte1:
Might want to change your comments to include your biome changes.

Issue #1008, opened 2017-04-05
[1.10.2] Seed oil inconsistency


JasonMcRay:
In 1.7.10 Seed oil that was created with GT machines was the same as forestry one. In 1.10 Forestry did changed some internal names and now there are 2 types of seed oil, that are not compatible with each other.

GT is adding `<liquid:seedoil>` but Forestry has `<liquid:seed.oil>`

I talked to @Blood-Asp and as long as there is no exact reason why it is different, it should be the same. So GT machines (like fluid extractor) should give the `<liquid:seed.oil>` from extracting fluid from seeds.

PR #1011, opened 2017-04-06
Downgraded forge and better seedoil compat

ghost:
Changed seedoil to seed.oil for better forestry compat as requested here #1008.
Updated IC2.
Downgraded Forge and mappings.

I downgraded forge and mappings because IC2 did not work with the newer versions because of a biome related method called when generating a new world.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ archivesBaseName = "gregtech"
 
 minecraft {
     version = "${config.minecraft.version}-${config.forge.version}"
-	mappings = "snapshot_20161111"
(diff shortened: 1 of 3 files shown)


Dragon2488:
So, you updated mappings and haven't checked you changes working before sending previous pull request? I merged it without warranties, but now it doesn't work. I will test your pull in work tommorow, enough blind hope in people.

Also I will beter wait Chocohead to update ic rather than downgrading. It shouldn't be very long.

Also your fluid name change will result missing fluid in all existing worlds. So, no.

draknyte1:
You can always revert a merge.

PR #1014, opened 2017-04-08
Delete some unused staff

Exidex:
Delete some unused methods and classes


draknyte1:
You can't remove deprecated stuff in materials.java. 
While not used by Gregtech it may be used by people who support all gregtech versions for addons. (Me)

Blood-Asp:
Actually now is the right time to remove that stuff. It was only left to not break existing stuff, but currently there does not exists anything that uses it. Cleaning that up now will have it cleaner later should addons actually move.

draknyte1:
I've been informed this was 1.10, mobile doesn't dictate which is which. In such a case, that's fine.

PR #1018, opened 2017-04-10
NEI Large Boiler Fuel Tab

JohannesGaessler:
Expanded the NEI api to allow for custom descriptions, added liquid fuels to the newly created Large Boiler tab

Instead of the default information like total EU or EU usage it is now
possible to provide custom Strings to describe a recipe.

The Large Boiler NEI tab explicitly shows burn times for fuels depending
on the boiler used.
Any liquid fuel added to Diesel Fuels or Semifluid Fuels will
automatically be added to Large Boiler Fuels.


JohannesGaessler:
I'm currently looking for a way to get all items with a fuel Value >0 but so far I didn't find one.

Blood-Asp:
For the solid fuels: There is one step where GT is looping through all items. There you should be able to search for all fuels. Also, there is a default furnace fuels tab in NEI. But no idea how it is build. You should at last also add the thaumcraft alumentum as quite a few are using it if TC is added.

PR #1025, opened 2017-04-15
Steam production improvement / Tooltip improvement / Pollution adjustment

JohannesGaessler:
Machine descriptions can now consist of more than one String.
Machine tooltips are now more descriptive, especially in regards to steam production / consumption and pollution.

Pollution values increased for Gas Turbines and Diesel Generators.

Large Boilers can now be throttled down.

I'm not entirely sure whether my interpretation of the Diesel Engine source code is correct, please check!


Techlone:
@JohannesGaessler can you write examples (a table/grid would be perfect) how to change a burn rate of coal and a steam output per coal in Bronze Large Boiler for different inergated circuit values?
By default is 40 ticks and 32000L

JohannesGaessler:
The Steam rate is reduced by 1000L/s per circuit config.
The fuel burn rate is adjusted so that the total amount of steam per fuel is unchanged.

Example:
A Large Steel Boiler produces 24000L of Steam per second normally.
By inserting a circuit with config=12 you can have the boiler produce 24000L - 12 * 1000L = 12000L per second with half the normal fuel consumption

Techlone:
Oh I see they just reduce a steam output per second and also proportionally a fuel consumtion.
It's a useful behavior thanx

SapientMC:
@JohannesGaessler Cool idea. Thanx a lot.

Blood-Asp:
Rising the pollution produced by diesel gen and gas turbine 3x is too much imo. I see that is so that the smallest calculation unit is 1 pullution per tick and multiples of that, but maybe make the base to 1 pullution per second so the old balancing is mostly maintained.

JohannesGaessler:
If you want to run a basic steam turbine with coal boilers you already get 100 Pollution per second.
Gas (20) and Diesel (40) is still nothing compared to that.

Also in my opinion it doesn't make any sense to have the pollution suddenly jump from 70 (HV Diesel) to 300 (Diesel Engine), or from 35 (HV Gas) to 160 (Gas Turbine).
With the changes I proposed the pollution per EU would consistently halve with each tier.

Also the old formula for pollution is kind of nebulous.
When I calculated Pollution per second my results differed from the ones in the documentation by over 5% for HV Machines.

JohannesGaessler:
On another note, isn't the pollution you get from the Charcoal Pit also kind of low?
I'd imagine that it'd create about as much pollution as the Pyrolyse Oven.

src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_BasicBatteryBuffer.java
```diff
@@ -43,7 +43,7 @@ public GT_MetaTileEntity_BasicBatteryBuffer(String aName, int aTier, String[] aD
 
     @Override
     public String[] getDescription() {
-    	String[] desc = new String[mDescription.length];
+    	String[] desc = new String[mDescription.length + 1];
```
Techlone: @JohannesGaessler oh come on, a minus should be here

Techlone: @codewarrior0 what's wrong?

Techlone: @JohannesGaessler @codewarrior0 oh excuse me I understood :) I thought that is the line with an assignment below

OvermindDL1: > @codewarrior0 what's wrong?

Because `- 1` would cause a buffer overflow, and crash.

Techlone:
GT log with this PR https://pastebin.com/77hMZE3H
Too many NullPointerException. I tried find out what's wrong with `mDescription` and why it's `null` but I didn't get it.

src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_Maintenance.java
```diff
@@ -46,10 +46,26 @@ public GT_MetaTileEntity_Hatch_Maintenance(String aName, int aTier, String aDesc
...
+
     @Override
     public String[] getDescription() {
-    	if(mAuto)return new String[]{mDescription,"4 Ducttape, 2 Lubricant Cells","4 Steel Screws, 2 Adv Circuits","For each autorepair"};
-        return new String[]{mDescription, "Cannot be shared between Multiblocks!"};
+    	if (mAuto) {
+    		String[] desc = new String[mDescription.length + 3];
+    		System.arraycopy(mDescription, 0, desc, 0, mDescription.length);
+    		desc[mDescription.length] = "4 Ducttape, 2 Lubricant Cells";
+    		desc[mDescription.length + 1] = "4 Ducttape, 2 Lubricant Cells";
```
Techlone: A double

JohannesGaessler: What do you mean by that?

Techlone:
Well if do not dig into the problem, I changed `mDescription` assignment to this
`mDescription = aDescription == null ? new String[0] : aDescription;` or
`mDescription = aDescription == null ? new String[0] : new String[]{aDescription};`
in `GT_MetaTileEntity_TieredMachineBlock` constructors (total four assignments) and now it works.

src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_Maintenance.java
Techlone: two times "4 Ducttape, 2 Lubricant Cells"

draknyte1:
Does this maintain backwards compatibility?
Can't really compile and test while at work.

JohannesGaessler:
There should only be a problem if you were changing the machine description after calling the constructor.

I'll probably refactor the String array and  add a deprecated field mDescription to maintain the signature.

Blood-Asp:
So many changed classes. I hope that i did not miss anything important looking through the changes.

Blood-Asp:
Guess i will merge this including the changed pollution values and test how they will actually act in world.

PR #1026, opened 2017-04-16
Fix Lighter behaviour

Exidex:
Lighter behaviour, build script and other fixes.

### DIFF
--- a/src/main/java/gregtech/common/blocks/GT_Block_Reinforced.java
+++ b/src/main/java/gregtech/common/blocks/GT_Block_Reinforced.java
@@ -12,18 +12,24 @@ import gregtech.api.objects.MaterialStack;
 import gregtech.api.util.GT_LanguageManager;
 import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_OreDictUnificator;
+import net.minecraft.block.Block;
 import net.minecraft.block.SoundType;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.renderer.texture.TextureAtlasSprite;
 import net.minecraft.client.renderer.texture.TextureMap;
 import net.minecraft.creativetab.CreativeTabs;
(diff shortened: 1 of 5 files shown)


src/main/java/gregtech/common/blocks/GT_Block_Reinforced.java
```diff
@@ -180,6 +187,62 @@ public void dropBlockAsItemWithChance(World worldIn, BlockPos pos, IBlockState s
...
+            EntityTNTPrimed entitytntprimed = new EntityTNTPrimed(world, (double)((float)pos.getX() + 0.5F), (double)pos.getY(), (double)((float)pos.getZ() + 0.5F), explosion.getExplosivePlacedBy());
+            entitytntprimed.setFuse(world.rand.nextInt(entitytntprimed.getFuse() / 4) + entitytntprimed.getFuse() / 8);
+            world.spawnEntityInWorld(entitytntprimed);
+        }
+        super.onBlockExploded(world, pos, explosion);
+    }
+
+    @Override
+    public boolean onBlockActivated(World worldIn, BlockPos pos, IBlockState state, EntityPlayer playerIn, EnumHand hand, @Nullable ItemStack heldItem, EnumFacing side, float hitX, float hitY, float hitZ) {
+        if (heldItem != null && (heldItem.getItem() == Items.FLINT_AND_STEEL || heldItem.getItem() == Items.FIRE_CHARGE) && state.getValue(METADATA) == 5) {
```
MauveCloud, line 230: The original summary of the PR was to fix the lighter behavior.  Is there a reason this line doesn't check for the lighter as an option for igniting the fuse of a powder barrel?

Exidex: I thought about it but I did not see way to check if item has lighter behavior without adding method to GT_MetaBase_Item and 5.09 doesn't check either so I left it as is.

PR #1038, opened 2017-04-27
Fix eu consumption absence by AssemblyLine and some clean up

Techlone:
Assembly line doesn't consume energy at all because `mEUt` is positive.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_AssemblyLine.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_AssemblyLine.java
@@ -67,58 +67,61 @@ public class GT_MetaTileEntity_AssemblyLine
     }
 
     public boolean checkRecipe(ItemStack aStack) {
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_AssemblyLine.java
```diff
@@ -130,6 +133,9 @@ public boolean checkRecipe(ItemStack aStack) {
                 this.mMaxProgresstime /= 2;
             }
         }
+        if (this.mEUt > 0) {
+            this.mEUt = -this.mEUt;
```
Techlone, line 137: nobody noticed this lol

Blood-Asp: Pretty sure i did something about that shortly after adding the multiblock. Likely got overwritten in some later git change after that. Happend a few times.

Issue #994, opened 2017-03-30
[1.10.2] Crash with rftools builder quarry mode


SiblisCu:
The game crash to desktop  every time I load the world.
This happens while quarring the overworld.

Crash report:

[20:29:25] [Client thread/INFO] [STDOUT/]: [net.minecraft.init.Bootstrap:func_179870_a:560]: ---- Minecraft Crash Report ----

WARNING: coremods are present:
  Brandon's Core (BrandonsCore-1.10.2-2.1.4.92-universal.jar)
  IvToolkit (IvToolkit-1.3.2.1-1.10.jar)
  ForgelinPlugin (Forgelin-1.4.0.jar)
  LoadingPlugin (ResourceLoader-MC1.9.4-1.5.1.jar)
  BookshelfLoadingPlugin (Bookshelf-1.10.2-1.4.3.338.jar)
  CCLCorePlugin (CodeChickenLib-1.10.2-2.5.6.233-universal.jar)
  IC2core (industrialcraft-2-2.6.142-ex110.jar)
  CorePlugin (MrTJPCore-1.10.2-2.0.0.14-universal.jar)
  TransformerLoader (OpenComputers-MC1.10.2-1.6.2.7.jar)
  CoFH Loading Plugin (CoFHCore-1.10.2-4.1.0.155-universal.jar)
  TheBetweenlandsLoadingPlugin (TheBetweenlands-2.0.2-alpha-universal.jar)
  ChiselCorePlugin (Chisel-MC1.10.2-0.0.8.12.jar)
  NEICorePlugin (NotEnoughItems-1.10.2-2.1.3.201-universal.jar)
  CoreMod (Aroma1997Core-1.9.4-1.1.1.1.jar)
  Default Options (DefaultOptions_1.10.2-6.1.5.jar)
  EnderCorePlugin (EnderCore-1.10.2-0.4.1.65-beta.jar)
  SFRCore (SolarFluxReborn_1.10.2-2.13r.jar)
  LoadingPlugin (Quark-r1.1-70.jar)
  ArsMagica2-Preloader (ArsMagica2-1.10.2-1.5.0-17b.jar)
  Do not report to Forge! Remove FoamFixAPI (or replace with FoamFixAPI-Lawful) and try again. (foamfix-0.6.2-beta2-anarchy.jar)
  RebornCoreASM (RebornCore-1.10.2-2.13.5.136-universal.jar)
[...]

PR #1042, opened 2017-05-03
[1.10.2] Rework some blocks, delete METADATA property

Exidex:
Rewrite some blocks using properties and some minor tweaks


src/main/java/gregtech/api/enums/Materials.java
```diff
@@ -1570,25 +1571,6 @@ public Materials(int aMetaItemSubID, MaterialSet aIconSet, float aToolSpeed, int
...
-	 * This is for keeping compatibility with addons mods (Such as TinkersGregworks etc) that looped over the old materials enum
-	 */
-	@Deprecated
-	public static Materials valueOf(String aMaterialName) {
-		return getMaterialsMap().get(aMaterialName);
-	}
-
-	/**
-	 * This is for keeping compatibility with addons mods (Such as TinkersGregworks etc) that looped over the old materials enum
-	 */
```
JohannesGaessler: Are you sure we should just remove these methods? I don't know whether TinkersGregwork is still relevant but is there an actual benefit to removing the deprecated methods?

JohannesGaessler (commented):
I have trouble understanding the nature of your changes.
Please describe what you changed and why.

Exidex:
So instead of having one Property METADATA which is just a number, for example ores now have 2 properties which represent StoneType and Material and you can now access those properties of an ore without any manipulations on a some number from 0 to 15.

draknyte1:
I assume this is for 1.10.
Changing things like this for 1.7.10 isn't really a valid idea.

Exidex:
It is targeted at 1.10.2 branch.

src/main/java/gregtech/api/enums/Materials.java
Exidex: There is no addons for 1.10.2 for GT and these methods have their not deprecated analogs

draknyte1: I'll probably use them initially when I port to 1.10.2

Exidex: So should I revert it?

draknyte1: Leaving them does no damage. I would appreciate it.)

src/main/java/gregtech/common/blocks/GT_Block_GeneratedOres.java
```diff
@@ -34,33 +35,35 @@
...
-    public static void doOreThings() {
-        System.out.println("MATERIALS META OFFSET: " + MATERIALS_META_OFFSET);
-        System.out.println("MATERIALS/BLOCK: " + MATERIALS_PER_BLOCK);
-        sGeneratedBlocks = new IBlockState[GregTech_API.sGeneratedMaterials.length][];
-        sGeneratedSmallBlocks = new IBlockState[GregTech_API.sGeneratedMaterials.length][];
         Materials[] lastMats = new Materials[MATERIALS_PER_BLOCK];
         int length = 0;
-        for(Materials aMaterial : Materials.values()) {
-            if(aMaterial != null && aMaterial.mMetaItemSubID >= 0 && (aMaterial.mTypes & 0x08) != 0) {
+        for(Materials aMaterial : GregTech_API.sGeneratedMaterials) {
```
Exidex: This will fix the issue that was discussed in #994

KartAlexPet:
When will there be a new version with a solution?
The problem with rftools is still relevant.

Exidex:
@KartAlexPet GT5U for 1.10 is now under a major rewrite, some things are done, more to do. I don't think it will be ready at least for a couple months.

KartAlexPet:
I understood. Good luck.

PR #1048, opened 2017-05-06
[5.09.29] EBF recipe fix

JohannesGaessler:
When checking for Recipes the EBF currently makes a FluidStack[] from only the first FluidStack in its Input Hatches.
This has the effect of one fluid input being able to 'block' all others.

With the changes I made the EBF now considers all FluidStacks for finding a recipe.

Also added a string representation for FluidStacks.


codewarrior0:
This one tripped me up for a bit when I tried to make silicon wafers in the same EBF I use for steel.

But it looks like you have a stray commit on this branch?

draknyte1:
Don't have a PR for an EBF fix and try add features outside of that.
Move your Large Chem Reactor to it's own PR...

JohannesGaessler:
Sorry about the unrelated commit, it was not intentional.
I created a local branch for the Large Chemical Reactor and accidentally commited onto the wrong branch.

Issue #1059, opened 2017-05-17
Mistake in chemical formula of Diatomite


Antillar:
Diatomite has this formula: SiO28Fe2O3Al2O3
According to the recipe in Electrolyser, the formula should be (SiO2)8Fe2O3Al2O3, so it has 8 Si and 16 O in 8 silicium dusts instead of 1 Si and 28 O.

PR #1060, opened 2017-05-18
Material formula fix

JohannesGaessler:
Currently when checking whether brackets are to be put around one of the MaterialStacks in a Materials chemical formula the conditions are:
-Stack size > 1
-Direct stack submaterials > 1

However, this does not detect Materials like Flint, which has exactly one direct submaterial (SiliconDioxide).
For example currently the formula for Diatomite is SiO28Al2O3Fe2O3 instead of (SiO2)8Al2O3Fe2O3.
I added another case that adds brackets when the MaterialStack is made from exactly one compound material.

### DIFF
--- a/src/main/java/gregtech/api/objects/MaterialStack.java
+++ b/src/main/java/gregtech/api/objects/MaterialStack.java
@@ -1,7 +1,6 @@
 package gregtech.api.objects;
 
 import gregtech.api.enums.Materials;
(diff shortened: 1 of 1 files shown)


Blood-Asp:
Was this the fix for #1059 ?

JohannesGaessler:
Yes.

PR #1074, opened 2017-06-02
AdvMiner2 renamed; new tiers of OreDrillingPlant

Techlone:
*The miner is dead long live the miner.*

I removed a 5% chance to dig a block and multiplied a required time and an amount of a drilling fluid by 20. It's equivalent.

T1 is the old miner - steel casings/frames; radius is 3 chunks and MV+
T2 - titanium casings/frames; radius is 4 (+1) chunks and HV+
T3 - tungstensteel casings/frames; radius is 6 (+2) chunks and EV+
T4 - osmiridium casings/frames; radius is 9 (+3) chunks and IV+

ODP collects mining pipes after finish a mining. Also it digs an ore block if it's right below the mining tip.
And now ODP doesn't mine a layer where it has been placed; it begins dig below itself.

How they look (nothing new)
image

GT with this PR [link](https://drive.google.com/open?id=0B-9mIQRXNJUoQ2xGekNWMUJnbmM)
Link update. I merged the pollution bugfix.


JohannesGaessler:
There's a conflict between this PR and mine:
I've also added a Multiblock, the Large Chemical Reactor.
To this end I've added two casings in GT_Block_Casings4 at IDs 14 and 15, which overlaps with the id you're using for the osmiridium casing.

IDs 4 and 5 do not seem to be in use at the moment.
Unless there's a problem with using an ID that was previously used for another casing I suggest that one of us switches to those IDs.

Techlone:
@JohannesGaessler hmm did you add two blocks? Can you use these 4 and 5 IDs please?
If I will change then a single ID still will be stand empty but you can perfectly fill both.

Antifluxfield:
Why not add some new tiers for the Oil Drilling Rig, which can extract oils in more chunks (in the same oil field)?
For example:
T1: the old oil drill, steel casings/frames; radius is 1 chunk (1 * 1 chunk);
T2: titanium casings/frames; radius is 2 chunks (3 * 3 chunks), 4 of this should be enough to cover each oil field;
T3: tungstensteel casings/frames; radius is 4 chunks (7 * 7 chunks), 1 is enough for each 6 * 6 oil field.

Techlone:
@antifluxfield cuz tiers of OilDrillingRig is another idea, don't you think?

JohannesGaessler:
I changed my casing block IDs.
There should be no conflict now.

Techlone:
Merged with `unstable`, fixed conflicts, uploaded new jar file (see the head post).

PR #1077, opened 2017-06-04
Wire/Pipe tweak

Technus:


Technus:
Might still require getting changed pipe textures.

PR #1083, opened 2017-06-08
More fixes for the world gen

Techlone:
The issue is that `this.mX` and `this.mX` can be negative (surprise).
Therefore checks [here](https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/common/GT_Worldgenerator.java#L95-L98) are not fully.
Also `getVeinCenterCoordinate` gets you a right coordinate of a center chunk and checks `(X / 16 - 1) % 3 == 0` no need more.
Plus some IDEA's formatting.


Techlone:
And tons of merging commits of course

PR #1088, opened 2017-06-13
Reworking DataOrb and DataStick

Techlone:
Now DataOrb and DataStick are not circuits.
Changed a DataOrb recipe (includes a Nanoprocessor) ...
image
and a DataStick recipe (includes an Integrated Processor).
image

Added a true Crystalprocessor Assembly.
image

The old Crystalprocessor Assembly renamed to Ultimate Crystalcomputer.
image

[Link](https://drive.google.com/open?id=0B-9mIQRXNJUoalRreWpHRENxWEE) for a testing.


Georggi:
IDK why this is needed.... Their placement was fine in my opinion...

Blood-Asp:
The data storage are not really circuits and die orb had to be usable earlier for replication so it got a second recipe and was jumped over in the circuit progression.

So this change has a reason.

PR #1100, opened 2017-06-17
[5.09.30] Oil Refining / Rubber Update

JohannesGaessler:
As a Followup to my Chemistry Update, where I had added more complex recipes for making plastics, I have now also updated how to get the raw Materials for plastic production.

This mainly affects how Oil Cracking works.
Unlike before, Oil Cracking is now an important step in the production of plastics.
Essentially, the process now involves taking a material that consists of long hydrocarbon chains (Naphtha, Diesel, etc.), and cracking it with either Hydrogen or Steam to get shorter hydrocarbon chains useable for plastic production (Ethylene, Propene, etc.).
Instead of an Oil Cracker, a Chemical Reactor can also be used, albeit only at 80% efficiency.

There are two ways of cracking oil: Hydrocracking and Steamcracking.

**Hydrocracking** requires Hydrogen and is suitable for processing relatively heavy compounds (Heavy Fuel, Light Fuel) into lighter compunds (Naphtha).
Notably, Hydrocracking produces very little byproducts but also only produces saturated hydrocarbons which cannot be directly used to make plastic.

**Steamcracking** requires Steam and is suitable for turning relatively light hydrocarbons (Naphtha) into Olefins, the sort of reactive, unsaturated hydrocarbons that can be used to make plastic.
Each time a fluid is Steamcracked, a small amount of oil is turned into elemental Carbon.
Also, because of the reactive nature of olefins, some of the hydrocarbons will recombine into longer chains, meaning that a small part of the cracked fluid might actually be heavier than the input fluid.

Oil Cracking can be done at three different levels of Severity for each of the two methods, resulting in a total of six different recipes per input.
A higher severity generally means that the input is split into shorter chains.
All GT Oil products and all GT hydrocarbons can be cracked.

Since the amounts of different Materials you're supposed to be getting from Oil Cracking is very high, I changed how the Distillation Tower works.
The Distillation Tower can now have up to 11 Fluid outputs and 1 Solid output, with a variable height ranging from 3 to 11 blocks.
Essentially, fluids now have a set height, at which they can be extracted from the Distillation Tower (the number in the NEI GUI), and for each distillation product that you want to recover you need to build one extra layer.

I retired Glyceryl Trinitrate as the Material needed for Nitro Diesel and instead added Tetranitromethane. (produced mostly from Nitric Acid)

I added more two more Rubber types: Silicone Rubber (rename of Poly(dimethyl)siloxane), and Styrene-Butadiene Rubber.
Both of these Rubber types are of equal quality and are produced synthetically (though with different starting materials).
There is also a synthetic recipe for normal Rubber.
The new rubber types are needed for EV+ recipes, and can also be used in low-tech applications instead of regular Rubber.

Particularly, the new Rubber types are needed for Cable production.

Cables up to HV can be produced like before, although now there is an alternative recipe that requires a dielectric (Polydimethylsiloxane or Polyvinyl Chloride) and synthetic Rubber (Silicone Rubber or SBR) to insulate the wire (Total material cost half that of Rubber-only insulation)

EV+ Cables mandatorily require a dielectric and synthetic rubber, and also a shielding layer.
The shielding layer can be made either from the wire material (+50% cost), or from Polyphenylene Sulfide, a new plastic with the ability to conduct electricity.

Instrumentally, I increased the amount of inputs of the Assembler from 2 to 6.
Originally I planned to also add recipes that allow the assembler-based production of machine components (Motor, Conveyor Module, etc.), but I scrapped those plans.

Implemented Polystyrene, though it doesn't have a use as of yet.

To try and make things consistent I reblanced the burn value of hydrocarbons, most notably:
Methane:   45000 -> 108000
Hydrogen: 15000 ->   20000
Ethanol:   128000 -> 148000

I have added a new recipe for Glyceryl Trinitrate.
It is now produced from Glyerol (Bio Diesel byproduct) and Nitration Mixture.
You now need more glyceryl Trinitrate per Dynamite, but in return Dynamite is now actually useable in the Implosion Compressor.

I have changed the recipe for Sodium Persulfate.
It is now made from Salt and Sulfuric Acid.

Lastly I added Bio Diesel, which can be produced from mainly Seed Oil/Fish Oil, some Methanol/Ethanol, and small amounts of Sodium Hydroxide.
To allow for scaleable Seed Oil production without having to rely on other mods I added a new Crop: Rape.


draknyte1:
Two things:
1.) I hope the original constructors for recipe adds still exist. And just call the new ones. 
2.) infinite fuel loops with methane. Fish power for life.

JohannesGaessler:
@draknyte1 
1.) The only thing that would require anyone to change their code is the Distillation Tower. Previously the order in which you specified the outputs was from top to bottom, now it is bottom to top so that it aligns with the NEI recipe.
2.) I'm fairly confident that with only GT it is not possible to build a perpetuum mobile. If you play with Galacticraft you can build one with the GC Electrolyzer, but that was already possible before I changed anything.

draknyte1:
> 1.) The only thing that would require anyone to change their code is the Distillation Tower. Previously the order in which you specified the outputs was from top to bottom, now it is bottom to top so that it aligns with the NEI recipe.

So, the constructor has changed for adding recipes?
If so, Re-Add/Leave the old one and in RecipeAdder make it use the new one.

JohannesGaessler:
What has actually changed is how the Distillation Tower determines where to output it's fluids.
The Output Hatches now have to be at the correct heights (1 Output hatch at each y coordinate).

draknyte1:
Well, that doesn't effect me then. :D
Just making sure the few recipes I add don't need special compatibility, because it's a pain to add via reflection.

Issue #1098, opened 2017-06-16
Machine texture is bugged in inventory


Antillar:
I have a video with this issue because it is hard to describe it with words and/or screenshots
https://www.youtube.com/watch?v=ZkQlxd8dITA&feature=youtu.be

Technus:
Might be related to dynamic hitboxes, probably missing override for inventory hitbox?

Technus:
https://github.com/Blood-Asp/GT5-Unofficial/pull/1118/commits/850f30a9d7a55608e62cc8d4ca972b03b9045e66

PR #1118, opened 2017-06-25
Assembly line recipe NEI support, and #1098 bugfix

Technus:
Might require adding/changing some Translations in Lang file.


Blood-Asp:
This PR now has conflicts after the last merges.

JohannesGaessler:
@Blood-Asp Do you want me to merge it? Since I made the changes that now cause conflicts I should be able to resolve them relatively easily.

Blood-Asp:
that would be good too. i'm currently on another project and merging would mean to switch some stuff back in the middle...

PR #1126, opened 2017-06-28
Trying to fix PE polymerizing recipe

Antifluxfield:
These recipes collided with the PE polymerizing recipe with oxygen.
I add an Integrated Circuit to distinguish them. Do you have any better idea?


JohannesGaessler:
I don't think there's a better approach other than adding integrated circuits.
Is there a reason why you disabled the recipe with only cells?

Antifluxfield:
That recipe collides with the PE recipe in the Large Chemical Reactor.

Antifluxfield:
We can fix the third recipe with that simple boolean. But I think we'd better add those methods to sustain their consistency.
If you think adding those things is OK, then all the work is done.

Antifluxfield:
Revert last commits; now using another individual method to add recipes only to the basic Chemical Reactors, trying to evade some potential conflicts.

PR #1128, opened 2017-06-29
Removed divider that already applied in Gt_UndergroundOil.

Dimach:


JohannesGaessler:
So the way I understand it, currently the fluid amount for the Oil Drilling Rig is divided two times:
Once in the multiblock, and once in GT_UndergroundOil.
Because of this the Oil Drilling Rig produces 5000 times less Oil than intended (0 because of rounding).
Is my interpretation correct?

Also, if possible please provide more context for future PRs.

SapientMC:
@JohannesGaessler Current v30 broke all OilRigs (which works fine on v29). Hope this PR will fix it.

Dimach:
@JohannesGaessler i removed divide by 5000 from OilRig that already applied in GT_UndergroundOil,  because when Technus moved methods of getting oil from GT_Utility they forget remove divider from OilRig. What more context you want?

Antifluxfield:
Well, you forget to fix the Seismic Prospector:
https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/api/util/GT_Utility.java#L2053

Blood-Asp:
And the scanner?

Dimach:
I fix it if you want. Oil rig is priority thing, scanners not.

Antifluxfield:
Scanner is OK. Delete that "/ 5000" is enough for the SP.

PR #1129, opened 2017-06-29
Player detector fix

Antifluxfield:
Revert #1067, that's something I've done wrong.
I make the BaseMetaPipeEntity able to record the placer of the Player Detector, so generally the Detector will work as well as it is put on BaseMetaTileEntity.

Or should we change the ICoverable interface, to make the tiles able to record the placer of the cover on each side?

### DIFF
--- a/src/main/java/gregtech/common/covers/GT_Cover_PlayerDetector.java
+++ b/src/main/java/gregtech/common/covers/GT_Cover_PlayerDetector.java
@@ -57,8 +57,8 @@ public class GT_Cover_PlayerDetector extends GT_CoverBehavior {
         if(aCoverVariable <0){aCoverVariable = 2;}
         switch(aCoverVariable) {
             case 0: GT_Utility.sendChatToPlayer(aPlayer, trans("068", "Emit if any Player is close")); break;
-            case 1: GT_Utility.sendChatToPlayer(aPlayer, trans("069", "Emit if other Player is close")); break;
-            case 2: GT_Utility.sendChatToPlayer(aPlayer, trans("070", "Emit if you are close")); break;
+            case 1: GT_Utility.sendChatToPlayer(aPlayer, trans("070", "Emit if you are close")); break;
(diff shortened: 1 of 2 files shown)


JohannesGaessler:
When you place a Player Detector Cover on a Pipe, and then someone else puts another cover onto the same pipe, does that not override the first cover?
Would it be reasonably possible to record who placed the Pipe instead of who placed the cover?

Antifluxfield:
I find that they will record who place the Pipe even without that line...
Well, now the second cover won't override the first one. They will all check who placed the Pipe, not the cover.

src/main/java/gregtech/common/covers/GT_Cover_PlayerDetector.java
```diff
@@ -57,8 +57,8 @@ public int onCoverScrewdriverclick(byte aSide, int aCoverID, int aCoverVariable,
         if(aCoverVariable <0){aCoverVariable = 2;}
         switch(aCoverVariable) {
             case 0: GT_Utility.sendChatToPlayer(aPlayer, trans("068", "Emit if any Player is close")); break;
-            case 1: GT_Utility.sendChatToPlayer(aPlayer, trans("069", "Emit if other Player is close")); break;
-            case 2: GT_Utility.sendChatToPlayer(aPlayer, trans("070", "Emit if you are close")); break;
+            case 1: GT_Utility.sendChatToPlayer(aPlayer, trans("070", "Emit if you are close")); break;
```
Blood-Asp, line 60: https://github.com/Antifluxfield/GT5-Unofficial/commit/c42c2facb1379c30e5642738bf1dd42cd9f16282#diff-597697fb4fcc15ac627355ef6b90bed2 what now?

Blood-Asp: ok, reading it better i see that you noticed your own mistake.

PR #1132, opened 2017-07-01
Item Distributor

JohannesGaessler:
I added a new Machine: The Item Distributor.
Like the Chest Buffer it buffers up to 27 Item Stacks, but instead of outputting everything to a single Side it distributes its input between up to five sides.
The amount of items per side can be set to any value between 0 and 127.
The purpose of the Item Distributor is to more easily distribute items between machines, and to split stacks.

To make the Item Distributor work I changed GT_Utility.moveOneItemStack, specifically in regards to the parameters aMaxMoveAtOnce and aMinMoveAtOnce.
Against my expectation these parameters did not actually limit the amount of items moved by this method.
Instead they limited the amount of items moved to a single output slot.

For example a stack of 9 items with aMaxMoveAtOnce=2, aMinMoveAtOnce=1 would be moved as follows:
9 -> 2 + 2 + 2 + 2 + 1

This behavior is almost identical to the one you get by manipulating aMaxTargetStackSize and aMinTargetStackSize.
The only difference is that with aMaxMoveAtOnce, aMinMoveAtOnce the shape of the output depends on the shape of the input.
For example with aMaxMoveAtOnce=2, aMinMoveAtOnce=1:
5 + 3 + 8 -> 6 + 5 + 3 + 2
16 -> 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2

In GT5U until now aMaxMoveAtOnce and aMinMoveAtOnce were never actually used so presumably this behaviour is not intended.
I changed it so that these parameters now actually limit the amount of items moved per method call.

The above pic shows the current Texture for the Item Distributor.
I'm not sure if I like it so any feedback is welcome.

Apart from the addition of the Item Distributor I have also made the description of Automation Machines more descriptive.


draknyte1 (changes requested):
Stop storing the .svg files.

Antillar:
I suggest adding the option to "not to output less than". Many chemical compound dusts in various quantities need to be processed in centrifuge or electrolyzer. Attempts to automate this require one machine per 1 dust, which is too much in terms of space taken by this setup.
If this block could hold items placed in it and output them once there are more than or equal to the defined number at once, so this group of items can be processed immediately, this would solve this problem.

JohannesGaessler:
@draknyte1 
The purpose of a version management system like Git is that you get continuous backups of development resources.
This is primarily source code, but also design documents or files from which other resources are generated.
As such I would argue that svg files do belong into the Git repository.

@Antillar
I don't think that change would have the desired effect.
The Item Distributor does not make a difference between items, therefore you wouldn't be able to effectively use it to regulate stack size per item, even if you defined a minimum stack size to move.
An Item Regulator would probably be more suitable.

draknyte1:
Nobody else stores theirs, no reason you need to either. You're not special, conform to the standard set in place. It's a source code repository.

JohannesGaessler:
This is not about what is the current practice but what is the better practice.
I put my svg files into the GT5U repository because it's what I would want everyone else to do, too.
The more people put their svg files into the repository, the easier it will be for everyone to recycle textures.

Especially recoloring a texture is extremely easy with an svg file.
Considering how many similar textures GT has (Casings, for instance), at least I would want to have access to as many svg files as possible because it would mean less work for me.

draknyte1:
People have had the choice and choose not to. Texture development work is committed when finalised and until that point you can keep it locally. I'm going to assume we all just use .Pngs. Recolouring images is easy if you're using anything other than paint. I personally think it's useless clutter you're adding to the repo and if others thought it a good idea they'd already be doing it. As it is we recycle textures without your method. 

Store them elsewhere please.

Blood-Asp:
I did not really think about storing the svg files... I don't think the idea is bad. Unlikely i will use them and it is a bit late, but i see nothing wrong with it. It is not in the way, they are small, they can be usefull. So why not?

draknyte1:
It's bloat that won't get used by anyone. 
Let's just store every file we've ever made for development here or let's just keep it to the source code for project builds. The latter option makes more sense and keeps the repo cleaner.

Dragon2488 (changes requested):
Just remove svg. It is useless. Svgs would be useful only if all of machines will have them.

OvermindDL1:
Does it only spit out equal amounts per side?  Will it not spit out amounts if a side is blocked?  Etc...

JohannesGaessler:
@OvermindDL1 
>The amount of items per side can vary from 0 to 127.

The items can be split unevenly.
By default, all sides receive zero items.
Items will be dropped into the world if a side with a weight greater than zero has air in front.
Items will not be put out to blocked sides.
In that case the Item Distributor will simply stop working until it can output the item.

OvermindDL1:
> The items can be split unevenly.

Any chance to add a button that enforces equal distributions among all sides with the ability to remove a side from consideration by putting a cover on it (turning that side 'off' basically)?

It would be nice to be able to put out equal and only equal amounts among all sides, right now I have to do something like that via...convolution.

JohannesGaessler:
For all sides to receive equal amounts of items all you have to do is set the weight for each relevant side to 1.
That means that you have to click each relevant side with the screwdriver once.
Perhaps there has been a misunderstanding about how the Item Distributor works because to me that doesn't seem very convoluted.

When I said that the amount of items per side can vary from 0 to 127, what I meant is that the player can set a value in that range for each side, not that there is a fluctuation.

src/main/java/gregtech/loaders/preload/GT_Loader_MetaTileEntities.java
```diff
@@ -1550,6 +1550,28 @@ private static void run4() {
...
+        GT_ModHandler.addCraftingRecipe(ItemList.Automation_ItemDistributor_LuV.get(1L, new Object[0]), bitsd, new Object[]{"XCX", "VMV", " V ", 'M', ItemList.Hull_LuV, 'V', ItemList.Conveyor_Module_LV, 'C', OreDictNames.craftingChest, 'X', OrePrefixes.circuit.get(Materials.Good)});
+        GT_ModHandler.addCraftingRecipe(ItemList.Automation_ItemDistributor_ZPM.get(1L, new Object[0]), bitsd, new Object[]{"XCX", "VMV", " V ", 'M', ItemList.Hull_ZPM, 'V', ItemList.Conveyor_Module_LV, 'C', OreDictNames.craftingChest, 'X', OrePrefixes.circuit.get(Materials.Good)});
+        GT_ModHandler.addCraftingRecipe(ItemList.Automation_ItemDistributor_UV.get(1L,  new Object[0]), bitsd, new Object[]{"XCX", "VMV", " V ", 'M', ItemList.Hull_UV,  'V', ItemList.Conveyor_Module_LV, 'C', OreDictNames.craftingChest, 'X', OrePrefixes.circuit.get(Materials.Good)});
+        GT_ModHandler.addCraftingRecipe(ItemList.Automation_ItemDistributor_MAX.get(1L, new Object[0]), bitsd, new Object[]{"XCX", "VMV", " V ", 'M', ItemList.Hull_MAX, 'V', ItemList.Conveyor_Module_LV, 'C', OreDictNames.craftingChest, 'X', OrePrefixes.circuit.get(Materials.Good)});
```
Antifluxfield, line 1574: Shouldn't they use differnet tiers of Conveyor Module?

JohannesGaessler: They should;
I forgot to change the tiers from LV.

PR #1135, opened 2017-07-01
Fixed inventory misalignment when updating 5.09.30 -> 5.09.31

JohannesGaessler:
I added a new method to BaseMetaTileEntity that shifts the Inventory index according to the change in Input/Output slots.
This fixes FluidStacks ending up in ItemStack slots.

Instrumentally I have also added a new version ID that considers the GT subversion.

### DIFF
--- a/src/main/java/gregtech/GT_Mod.java
+++ b/src/main/java/gregtech/GT_Mod.java
@@ -60,7 +60,8 @@ import java.util.regex.Pattern;
 
 @Mod(modid = "gregtech", name = "GregTech", version = "MC1710", useMetadata = false, dependencies = "required-after:IC2; after:Forestry; after:PFAAGeologica; after:Thaumcraft; after:Railcraft; after:appliedenergistics2; after:ThermalExpansion; after:TwilightForest; after:harvestcraft; after:magicalcrops; after:BuildCraft|Transport; after:BuildCraft|Silicon; after:BuildCraft|Factory; after:BuildCraft|Energy; after:BuildCraft|Core; after:BuildCraft|Builders; after:GalacticraftCore; after:GalacticraftMars; after:GalacticraftPlanets; after:ThermalExpansion|Transport; after:ThermalExpansion|Energy; after:ThermalExpansion|Factory; after:RedPowerCore; after:RedPowerBase; after:RedPowerMachine; after:RedPowerCompat; after:RedPowerWiring; after:RedPowerLogic; after:RedPowerLighting; after:RedPowerWorld; after:RedPowerControl; after:UndergroundBiomes;")
 public class GT_Mod implements IGT_Mod {
-    public static final int VERSION = 509;
+    public static final int VERSION = 509, SUBVERSION = 31;
+    public static final int TOTAL_VERSION = 1000 * VERSION + SUBVERSION;
     public static final int REQUIRED_IC2 = 624;
     @Mod.Instance("gregtech")
     public static GT_Mod instance;
--- a/src/main/java/gregtech/GT_Mod.java
+++ b/src/main/java/gregtech/GT_Mod.java
@@ -61,7 +61,7 @@ import java.util.regex.Pattern;
 @Mod(modid = "gregtech", name = "GregTech", version = "MC1710", useMetadata = false, dependencies = "required-after:IC2; after:Forestry; after:PFAAGeologica; after:Thaumcraft; after:Railcraft; after:appliedenergistics2; after:ThermalExpansion; after:TwilightForest; after:harvestcraft; after:magicalcrops; after:BuildCraft|Transport; after:BuildCraft|Silicon; after:BuildCraft|Factory; after:BuildCraft|Energy; after:BuildCraft|Core; after:BuildCraft|Builders; after:GalacticraftCore; after:GalacticraftMars; after:GalacticraftPlanets; after:ThermalExpansion|Transport; after:ThermalExpansion|Energy; after:ThermalExpansion|Factory; after:RedPowerCore; after:RedPowerBase; after:RedPowerMachine; after:RedPowerCompat; after:RedPowerWiring; after:RedPowerLogic; after:RedPowerLighting; after:RedPowerWorld; after:RedPowerControl; after:UndergroundBiomes;")
 public class GT_Mod implements IGT_Mod {
     public static final int VERSION = 509, SUBVERSION = 31;
-    public static final int TOTAL_VERSION = 1000 * VERSION + SUBVERSION;
+    public static final int TOTAL_VERSION = calculateTotalGTVersion(VERSION, SUBVERSION);
     public static final int REQUIRED_IC2 = 624;
     @Mod.Instance("gregtech")
     public static GT_Mod instance;
@@ -1163,4 +1163,11 @@ public class GT_Mod implements IGT_Mod {
         }
     }
     
+    public static int calculateTotalGTVersion(int minorVersion){
+    	return calculateTotalGTVersion(VERSION, minorVersion);
+    }
+
+    public static int calculateTotalGTVersion(int majorVersion, int minorVersion){
+    	return majorVersion * 1000 + minorVersion;
+    }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/gregtech/GT_Mod.java
```diff
@@ -60,7 +60,8 @@
 
 @Mod(modid = "gregtech", name = "GregTech", version = "MC1710", useMetadata = false, dependencies = "required-after:IC2; after:Forestry; after:PFAAGeologica; after:Thaumcraft; after:Railcraft; after:appliedenergistics2; after:ThermalExpansion; after:TwilightForest; after:harvestcraft; after:magicalcrops; after:BuildCraft|Transport; after:BuildCraft|Silicon; after:BuildCraft|Factory; after:BuildCraft|Energy; after:BuildCraft|Core; after:BuildCraft|Builders; after:GalacticraftCore; after:GalacticraftMars; after:GalacticraftPlanets; after:ThermalExpansion|Transport; after:ThermalExpansion|Energy; after:ThermalExpansion|Factory; after:RedPowerCore; after:RedPowerBase; after:RedPowerMachine; after:RedPowerCompat; after:RedPowerWiring; after:RedPowerLogic; after:RedPowerLighting; after:RedPowerWorld; after:RedPowerControl; after:UndergroundBiomes;")
 public class GT_Mod implements IGT_Mod {
-    public static final int VERSION = 509;
+    public static final int VERSION = 509, SUBVERSION = 31;
+    public static final int TOTAL_VERSION = 1000 * VERSION + SUBVERSION;
```
Techlone: maybe add a method for a calculation of this version?
and use it here and below in `shiftInventoryIndex` method

JohannesGaessler: Well, that would work in pretty much the same way.
Can you explain your thought process behind the suggestion?

Techlone: An example:
I'm a developer of a new addon for GT. I don't care how this version should be calculated I just want compare the current version and 35 version. Then I can write this
```
if(GT_Mod.TOTAL_VERSION > GT_Utils.calculateVersion(35) {
```
Sent the only one argument cuz 509 can be default so two methods are exist in fact
`calculateVersion(int majorVersion, int minorVersion)`
`calculateVersion(int minorVersion) //cals calculateVersion(GT_Mod.VERSION, minorVersion)`
Also the magic numbers in the code is an antipattern.

JohannesGaessler: When you say magic numbers, do you mean when I calculate the total version or when I shift the slot index?

Techlone: I meant 509blahblah [here](https://github.com/Blood-Asp/GT5-Unofficial/pull/1135/files/a6046ddfad81dca9b1c63d5f171b9ccd137f7288#diff-e87e5bc277764af62acec6845a86dff7R1921)

JohannesGaessler:
I decided to add the method for calculating the total version ID to GT_Mod instead of GT_Utility because it seemed to me like that calculation is strongly coupled with GT_Mod.

Techlone:
Well, it's not too important. The code became a little more readable it's more important for a supporting it by other people.

draknyte1:
How did inventory slots even get broken?

JohannesGaessler:
Gregtech machines have only a single array in which they store their ItemStacks: mInventory.
Inputs, Outputs, FluidStacks, etc. are all stored consecutively in this array.
When the machine is stored as NBT, the contents of mInventory are stored by their index.
Then, when the NBT data is read, the contents of mInventory are restored by their index.

When I changed the size of individual blocks within the structure of mInventory, i.e. I changed the number of input/output slots, the index in the NBT did no longer match up with the correct index in the new machines.
For example, NBT contents that belonged into one of the Fluid slots could end up in one of the Item slots.

draknyte1:
Only doing this change on particular Gregtech versions is a tad pointless because older versions won't be aware of the changes you've made and won't even be affected.

`if (nbtVersion < chemistryUpdateVersion) {}`
Is absolutely useless.

src/main/java/gregtech/api/metatileentity/BaseMetaTileEntity.java
```diff
@@ -1904,4 +1907,55 @@ public AxisAlignedBB getCollisionBoundingBoxFromPool(World aWorld, int aX, int a
...
+    		} else {
+    			return slotIndex;
+    		}
+    		newInputSize = 1;
+    		newOutputSize = 1;
+    	} else {
+        	return slotIndex;    		
+    	}
+    	int indexShift = 0;
+    	if (slotIndex >= GT_MetaTileEntity_BasicMachine.OTHER_SLOT_COUNT + oldInputSize) {
```
draknyte1, line 1953: You could just use the number 4 instead of making a private final int public.. Since It'll probably never change.

JohannesGaessler:
`nbtVersion < chemistryUpdateVersion` is true if the version, in which the machine was stored as NBT, is prior to 5.09.31.
This means that it is true when the number of machine slots between storing and reading the machine state has changed, and the indices of the machine's content need to be shifted.
Since the default value for an integer read from NBT is 0, the statement in question is evaluated to true when dealing with a save from before 5.09.31;
All GT versions prior to 5.09.31 effectively have a version of 0.0.0

src/main/java/gregtech/api/metatileentity/BaseMetaTileEntity.java
JohannesGaessler: I decided to reference the constant so that it's immediately apparent why 4 is being added.
I think hard-coding the value would make the code more difficult to understand.

draknyte1:
The version will NEVER be < 09.31 though...
Saves prior to .09.31 don't hold the NBT tags for lower versions anyway, so it's redundant code.

JohannesGaessler:
Yes it will be.
For a save from 09.30 or lower `aNBT.getInteger("nbtVersion")` will be evaluated to 0.
0 < 509031

draknyte1:
Why not just check if the tag exists? 
Instead of relying on a 0 value fallback.

JohannesGaessler:
Because I want this fix to be applicable for future GT versions.
It is conceivable that the number of machine slots could be changed again;
For instance DreammasterXXL has mentioned that he would like to increase the number of input slots on the mixer from 4 to 6.

Such changes could possibly be made in later GT versions (5.09.45 as an example), leaving 3 segments which have to be differentiated:
-Saves with version < 09.31
-Saves with a 09.31 >= version < 09.45
-Saves with 09.45 <= version

If only the existence of the tag was considered, the last two cases could not be properly differentiated; 
I would have to either add a new tag or rewrite the method logic to something resembling the current logic.

src/main/java/gregtech/api/metatileentity/BaseMetaTileEntity.java
draknyte1:
```
int indexShift = 0;
int OTHER_SLOT_COUNT = 4; //GT_MetaTileEntity_BasicMachine.OTHER_SLOT_COUNT
if (slotIndex >= OTHER_SLOT_COUNT + oldInputSize) {
```
And now you don't expose private vars.

draknyte1:
I just think in that case that it would be easier adding a mixer with 9 slots, as opposed to changing the current one. Why make it 6? It's such an obscure number.

In 60+ releases since .07.30, nobody has ever changed the slots around.
Considering how we're forever getting closer to End of Life, major changes should just roll into 5.10 where you won't go about breaking compatibility. This is how software should be developed, as opposed to just forever extending life of older products and continually breaking compatibility of existing saves.

JohannesGaessler:
There are big differences between GT5U and conventional, commercial software projects:
1.There are no clearly defined goals.
2.GT5U is free.
3.There is a varying number of developers

For regular games putting new content into a new version makes sense because you need to sell it in order to make a profit.
For a project like GT5U the methods of agile software development are more suitable.

Also, moving to 5.10 would be the ultimate move to break compatibility with existing worlds.
Essentially everyone plays GT5U on 1.7.10, and any modpack currently using the mod would have no way of updating.
If I was making changes to the fundamental workings of GT5U (like rewriting electricity), then I would agree that that should be done in a new iteration of Gregtech.
But since the changes I'm making are not of that nature, I don't agree that I should push my changes into 5.10.

SapientMC:
@JohannesGaessler What do you think about to add output slots to Arc Furnace (up to 6 or even 9)?

JohannesGaessler:
Could probably be done;
I'll look into it though during the next two weeks I likely won't have the time to do it.

PR #1140, opened 2017-07-05
Fix the Distillation Tower

Antifluxfield:
Now it has to strictly follow the description.


JohannesGaessler:
I disapprove of this change because it would make multiblock behavior inconsistent.
For any other multiblock you can have an arbitrary number of input/output hatches.
If you're used to that behavior from other multiblocks and then go and build a Distillation Tower with two outputs at the same height it will be hard to tell why the structure is wrong.

Antifluxfield:
@JohannesGaessler 
Well, sorry for that I have just mixed up the output restriction and the structure restriction... I've just find that you have already restrict the output to each layer.
I think you're point of view should be more reasonable, so I'm goning to cancel this PR.

PR #1142, opened 2017-07-06
Added LV and MV miners.

Dimach:
Added simple singleblock miners for game start (LV and MV age).
Miners dig ore only and require mining pipes (similar to multiblock miners), but without bonuses (produces ore blocks) and dont require drilling fluid. After finishing process collecting all pipes.
Inventory has 2 input slots (for pipes) and 2 output slots (for ore and pipes), but output slots works only as buffer and you must use item auto extraction to external storage.
Mechanics is next: to scan ore in one row, dig one ore block or move to other layer miner need one cycle.
When out of electricity don`t reset. You can use LV miner with ULV solar panel, they will works 4 times slower but correctly.
LV miner - cycle length 1 sec. Use 24 EU/T. Radius 8 blocks.
MV miner - cycle length 0.25 sec. Use 96 EU/T. Radius 24 blocks.


src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final ItemStack miningPipe = GT_ModHandler.getIC2Item("miningPipe", 0);
+    private static final Block miningPipeBlock = GT_Utility.getBlockFromStack(miningPipe);
+    private static final Block miningPipeTipBlock = GT_Utility.getBlockFromStack(GT_ModHandler.getIC2Item("miningPipeTip", 0));
+
+    int drillX, drillY, drillZ;
+    boolean isPickingPipes;
+    boolean waitMiningPipe;
+    static int[] radius = new int[]{8, 8, 24, 40}; //Miner radius per tier
+    static int[] speed = new int[]{20, 20, 5, 2}; //Miner cycle time per tier
+    static int[] energy = new int[]{24, 24, 96, 384}; //Miner energy consumption per tier
```
JohannesGaessler: I think radius, speed, and energy should be final.
Also, when you make a field final the naming convention changes as follows:
miningPipeBlock -> MINING_PIPE_BLOCK

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
```diff
@@ -0,0 +1,202 @@
...
+        return oreBlock.getDrops(getBaseMetaTileEntity().getWorld(), posX, posY, posZ, blockMeta, 1);
+    }
+
+    @Override
+    public void saveNBTData(NBTTagCompound aNBT) {
+        super.saveNBTData(aNBT);
+        aNBT.setBoolean("isPickingPipe", isPickingPipes);
+        aNBT.setInteger("drillX", drillX);
+        aNBT.setInteger("drillY", drillY);
+        aNBT.setInteger("drillZ", drillZ);
```
JohannesGaessler, line 188: Should waitMiningPipe not also be saved?

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
Techlone: >MINING_PIPE_BLOCK

Why?

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
Dimach: Nope, because after loading moveOneDown be called and do checking of input slots, not find pipes and set waitMiningPipe to true again.

JohannesGaessler (changes requested):
I think these single-block miners are poorly balanced against the multiblock Ore Drilling Plants added by @Techlone .
I think a cycle length of 1s for LV and 0.25s for MV is much too short.
By comparision the MV Ore Drilling Plant has a cycle length of 24s.

I suggest making cycle length and EU/t consistent with the Ore Drilling Plants, while keeping the smaller radius and lower output to compensate for the lower cost.

Specifically this would mean:
LV: 12EU/t for 48s per block
MV: 48EU/t for 24s per block

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
Dimach: I copy-paste that fields from multiblock miner. Maybe should also change it in its code?

Dimach:
@JohannesGaessler i think you not right about balance. That miner not added to replace drilling plant, it added to replace IC2 miners (base and advanced). And its balance is good now - they have same electricity usage (to mine one ore block), and a bit higher speed.
Also they added to simplify LV age, where  you need large amount of resources and the drilling plant is unavailable.
And cycle time 1 second not mean that he will dig one ore block per second, they use time to scan rows and move mining tip. Slowdown him to 48s cycle time is ridiculous!

Dimach:
Maybe slowdown miners in 2-4 times not ruin the balance, but not 48 times for sure.

JohannesGaessler:
>And cycle time 1 second not mean that he will dig one ore block per second, they use time to scan rows and move mining tip.

So the single-block miners mine every block in their area, not just the ore blocks?
I was assuming it would work like the Ore Drilling Plants, i.e. all non-ore blocks would be skipped.

SapientMC:
@JohannesGaessler Before offering to nerf new miners please thunk about how to feed them by energy in early game? LV Sodium Battery are enough to dig only 39 items (ore blocks / cobble under pipe / dusts from small ores). How to produce 24 eU/t far from base? 8V solar are more expensive than LV Miner itself and makes it work 3 times slower and only during day.

Dimach:
`So the single-block miners mine every block in their area, not just the ore blocks?`
Nope. Its not mine any blocks except ores. But they use time to find ore. For example to fing ore in one layer LV miner will use 18 cycles. MV miner - 50. When they go through empty layer they waste many time.

codewarrior0:
> Before offering to nerf new miners please thunk about how to feed them by energy in early game? LV Sodium Battery are enough to dig only 39 items

Centrifuge oil sands, distill to naphtha, burn in gas turbine. One cell of naphtha is worth ~125k EU, 2.5 times as much as a sodium battery.

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
draknyte1: > Why?

That's how Java naming conventions work?

JohannesGaessler:
In any case I would like the GT machines for auto-mining to be functionally consistent.
Therefore I would like the single-block miners to work like the Ore Drilling Plants and mine Ore blocks at a constant rate.

Dimach:
Okay, i think that would be interest mechanics and prevent fast skiping empty layers but if you want i can change this. But we still did not reach agreement about rates. You must consider that miner works without bonuses, and it must help players with mining ore at LV age. If it will be too slow players will dig by hands as before.

JohannesGaessler:
Regarding energy: It would be very simple to build 4 HP Coal Boilers and one LV Steam Turbine which would already be enough for the LV Miner.

SapientMC:
> Regarding energy: It would be very simple to build 4 HP Coal Boilers and one LV Steam Turbine which would already be enough for the LV Miner.

And do so on every single chunk you want to mine? BTW you will also need water source for boilers.

JohannesGaessler:
Alright, so previously I made the following mistake:
I based my suggestion for the cycle length on the speed at which Ores are mined when it should be based on how much ore is effectively put out.
After adjusting for the multiblock having 2.5 times the output of the basic machine you would get the following stats:
LV: 12EU/t, 20s per Ore Block
MV: 48EU/t, 10s per Ore Block

Dimach:
Well, its better but i still insist to leave energy usage as 24 eu/t and set speed to 10s/block. MV accordingly must have 96 eu/t and 5 s/block.

draknyte1:
I agree with Dimach.

Blood-Asp:
so many mails about this PR... 24EU/t, 10s/b and 96EU/t 5s/b sounds fine, it could even be slightly faster, the original speed of 1s/b certainly too fast. But i have a slight problem with the radius. 8b radius for LV is fine, but 24b for MV is a huge jump. considering how dense GT oreveins are, one can mine most GT oreveins with one miner setup. 
So i would lower MV range to 16.
For comparation, does someone have the values of the IC2 miner?

SapientMC:
@JohannesGaessler @Blood-Asp 
>LV: 12EU/t, 20s per Ore Block
MV: 48EU/t, 10s per Ore Block

What are you talking about? 12EU/t * 20s = 4800 eU per block.
IC Miner needs only ~1280 eU per block.
No one will use 4x more expensive variant.

Blood-Asp:
The IC2 miner is digging stone blocks too. And it uses a tool that looses durability. So even if it is better in EU, it is still much worse.

SapientMC:
LV Miner dig stone blocks too. Much smaller (only under mining pipe) but nonetheless.
IC miner uses electrical tool so it loose only energy not durability.

Blood-Asp:
And IC2 miner is much slower, and it has a maximum range of 6 blocks...

SapientMC:
So lets make GT Miner same speed as IC but also same (or only a little more expensive) energy cost.
Something like 8 eU/t and 8 sec per block for LV should be enough.

JohannesGaessler (approved):

JohannesGaessler:
10s/5s is fine with me, too.
Should the multiblocks be adjusted, too?
Specifically, should their mining speed and EU/t be doubled so it aligns with the basic machine?

@SapientMC 
3k EU is pretty much negligible.
Even at LV you get 24k EU from 1 Charcoal.
I think the more important factors are how quickly you can mine blocks and in how big of an area.

draknyte1:
> 20s/10s is fine with me, too.
> Should the multiblocks be adjusted, too?
> Specifically, should their mining speed and EU/t be doubled so it aligns with the basic machine?

I don't see a reason they should be changed.
Multiblocks should not directly correlate to single block usage/performance.

SapientMC:
@JohannesGaessler 
> Even at LV you get 24k EU from 1 Charcoal.

That's true. But outside base without infinite water you need something like RC Water Tank (3x3x3 multiblock) to produce water for small boiler.
Total: Miner, Small Steam Turbine, several Small Boilers and Water Tank are needed to dig ores from one chunk. And after that you need to move all this machines to another place.
Alternative: Just miner and ULV (or even 1V) solar panel on top of it. It's important to have enough energy buffer in Miner for one circle.

SapientMC:
May be it would be the best - to put specific parameters in config and set it how each one of us want.

draknyte1:
@SapientMC that's overkill. The current values you've put forward are fine by me.

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
Techlone: Well yes `static final` should be named an upper case. But I thought upper case names mean that they are constant and **value type** variables. MINING_PIPE_BLOCK is a reference type variable and because of this I asked.

SapientMC:
Summarizing:

Small Miner LV.
8 eU/t, 8 sec for operation (1280-1500 eU per block). 8m radius (area 17x17).
Internal buffer 2048 eU.

Small Miner MV.
32 eU/t, 4 sec for operation (2560-3000 eU per block). 16m radius (area 33x33).
Internal buffer 8192 eU.

Both miners works only with enough energy in internal buffer.

Dimach:
Okay, if all agree branch ready to merge.

Blood-Asp:
For me it looks ok.

JohannesGaessler:
@SapientMC 
>But outside base without infinite water you need something like RC Water Tank (3x3x3 multiblock) to produce water for small boiler.

Is there a reason why you can't just use a drain cover?

Also, I'm not saying that powering a Miner with Steam is the optimal strategy.
I'm specifically using Steam for my example because Dimach said the purpose of the Miner is to simplify the LV age and Steam will be what you start out with.

You only gain access to Solar Panels at the end of LV when you build an EBF, at which point you will probably also have liquid fuels.
Then all you'd need to power a LV machine away from base would be a LV Gas Turbine/Diesel Generator.
So I think getting power away from your base is not as hard as you make it out to be.

@Dimach 
I think that if you're gonna call the machines LV Miner and MV Miner then they should actually be LV and MV, not ULV and LV.

Dimach:
In game they called Base miner and Advanced miner. And tier of machine is chosen not by their consumption (for example macerators use 2 EU/T) but by voltage.

SapientMC:
@JohannesGaessler 
> Is there a reason why you can't just use a drain cover?

Yep. Finite water.

JohannesGaessler:
I just tested the Drain Cover again just to be sure but as far as I can tell it does give you infinite Water unless you have something else installed that prevents making a vanilla infinite Water source.

SapientMC:
CodeChickenCore.cfg
finiteWater=true

PR #1144, opened 2017-07-07
Fixed coloring of pyrolyse oven.

Dimach:
Before

After


draknyte1:
Why not fix casing texture 0 instead of making 111?

Dimach:
Its impossible to use 0 index. If machine block receive zero index to paint yourself they use default color instead.

PR #1147, opened 2017-07-08
Add new zpm and uv battery

SLiMylya:
add new battery

change recipe


draknyte1:
We're not changing the storage of the ultimate battery.

Techlone:
@draknyte1 what's a problem with it?

Blood-Asp:
Personally i would prefer to keep MAX Voltage only as transfer Voltage, not for storage.

SapientMC:
@Blood-Asp we can change MAX voltage to 2M eU/t

Blood-Asp:
I mean the highest Power tier should be Transfer only. Same as EV for IC2 alone.

SapientMC:
@Blood-Asp why? This will simply require a transformer at both ends of the line.

draknyte1:
It should stay as a max value. 
Why change it? I disapprove of this PR until the original values for max are restored.

draknyte1 (changes requested):
Restore Max voltage battery capacity.

Antillar:
What about adding the most expensive battery craft (currently requested Max) to be UV, but already charged? It would set new goals for players in the endgame. 
The recipe can be made more expensive then.

SLiMylya:
change

or delete her? this battery from achievement. Ultimate battery for the game, because the battery buffer with it can be automated

draknyte1:
Leave the old battery as is..

If you want a zpm battery, add a config option to make empty ZPM's craftable and rechargeable.

SLiMylya:
old battery on max capacity not  automated

draknyte1:
I don't understand sorry. 
MAX violate will, as blood suggests, be for power transfer and not storage.

SLiMylya:
Sorry, max capacity

src/main/java/gregtech/common/items/GT_MetaGenerated_Item_01.java
```diff
@@ -438,17 +438,16 @@ public GT_MetaGenerated_Item_01() {
...
 
-        ItemList.Energy_LapotronicOrb2.set(addItem(tLastID = 599, "Lapotronic Energy Orb Cluster", "", new Object[]{new TC_Aspects.TC_AspectStack(TC_Aspects.ELECTRUM, 16L), new TC_Aspects.TC_AspectStack(TC_Aspects.VITREUS, 16L), new TC_Aspects.TC_AspectStack(TC_Aspects.POTENTIA, 16L), OrePrefixes.battery.get(Materials.Ultimate)}));
+        ItemList.Energy_LapotronicOrb2.set(addItem(tLastID = 599, "Lapotronic Energy Orb Cluster", "Reusable battery", new Object[]{new TC_Aspects.TC_AspectStack(TC_Aspects.ELECTRUM, 16L), new TC_Aspects.TC_AspectStack(TC_Aspects.VITREUS, 16L), new TC_Aspects.TC_AspectStack(TC_Aspects.POTENTIA, 16L), OrePrefixes.battery.get(Materials.Ultimate)}));
         setElectricStats(32000 + tLastID, 1000000000L, GT_Values.V[6], 6L, -3L, true);
 
-        ItemList.ZPM2.set(addItem(tLastID = 605, "Ultimate Battery", "Fill this to win minecraft", new Object[]{new TC_Aspects.TC_AspectStack(TC_Aspects.ELECTRUM, 64L), new TC_Aspects.TC_AspectStack(TC_Aspects.VITREUS, 16L), new TC_Aspects.TC_AspectStack(TC_Aspects.POTENTIA, 64L)}));
-        setElectricStats(32000 + tLastID, Long.MAX_VALUE, GT_Values.V[8], 8L, -3L, true);
```
Dragon2488, line 451: Do not touch Ultimate Battery max charge. Or make this change optional.

Dragon2488 (changes requested):

src/main/java/gregtech/common/items/GT_MetaGenerated_Item_01.java
draknyte1: Easier to just not change it and make a UV tier battery that holds less. 

Then make a config option to make  zpm craftable and rechargeable

SLiMylya:
Create zpm, create UV capacity 100,000,000,000.and do not touch ultimate battery. I understand correctly?

draknyte1:
ZPM battery exists, it is single use. 
Just make config option to add recipe and allow it to be rechargeable.

SLiMylya:
updated pr

SLiMylya:
added config EnableZPMandUVBatteries default false

Blood-Asp:
From what i can see are the requested changes done. So re-rewiew?

PR #1149, opened 2017-07-09
Fix oreBasalticMineralSand properly being mapped to "ore" prefix

joakibj:
PFAAGeologica has the Basaltic Mineral Sand block oredicted as oreBasalticMineralSand.

As title, this fix maps the oreBasalticMineralSand to the "ore" prefix, instead of the "oreBasalt" prefix, yielding the wrong material "icMineralsand". Now it should yield the correct materials to be usable with GT.

### DIFF
--- a/src/main/java/gregtech/api/enums/OrePrefixes.java
+++ b/src/main/java/gregtech/api/enums/OrePrefixes.java
@@ -795,6 +795,7 @@ public enum OrePrefixes {
         for (OrePrefixes tPrefix : values())
             if (aOre.startsWith(tPrefix.toString())) {
                 if (tPrefix == oreNether && aOre.equals("oreNetherQuartz")) return ore;
+                if (tPrefix == oreBasalt && aOre.equals("oreBasalticMineralSand")) return ore;
                 return tPrefix;
             }
         return null;


Dream-Master:
Will this fix the issue that some GT Ore variants not have recipes  in macerators as an example ?
I dont use Geologica or is this a Gelogica specific fix?

joakibj:
Yes, it fixes all ores in the oredictionary as "oreBasalticMineralSand" specifically to generate the correct processing recipes. In this case directly smelt to iron, macerate to 2x Basaltic Mineral Sand dust + 1x Basalt Dust, hammer recipe and blast furnace iron doubling.

edit: this is a PFAAGeologica specific fix in practice, unless other mods adds this type of sand as well.

Dream-Master:
I add it in my 90+ ore mixes

draknyte1 (approved):
I see no reason this shouldn't be approved.

Technus:
same thing might happen when an ore has Nether in name, iirc it was patched

Technus:
and yes @Dream-Master  we need it

Dream-Master:
@Technus we have it in us gt fork allready.

PR #1153, opened 2017-07-11
Multiblock Fluid IO fix

JohannesGaessler:
Currently the way multiblocks handle Fluid inputs/outputs is very impractical and also in some cases inconsistent.
Specifically it is currently not possible to use one Large chemical Reactor to desulfurize several materials.

Input:
Currently GT_Recipe.findRecipe will return the first recipe for which the Input types are equal, completely ignoring stack sizes.
For example if you have 3 Fluid inputs in the Large Chemical Reactor that allow for two separate recipes only one of the recipes will ever be returned;
One recipe is effectively blocking the other.
I added an option to findRecipe that lets you get only recipes which can actually be executed.
So far the only multiblock that makes use of this option is the Large Chemical Reactor.

Output:
Currently when a recipe has only one Fluid output the multiblock will look for any Output Hatch to dump the fluid.
When a recipe has more than one Fluid output, however, there will be a strict mapping for the Fluid stacks:
Fluid Stack 1 -> Output Hatch 1, Fluid Stack 2 -> Output Hatch 2, etc.
This causes FluidStacks to be voided even if there are empty Output Hatches.
I made it so that multiblocks will always only void Fluids if there are no empty Output Hatches available.


SapientMC:
Think this option `to findRecipe that lets you get only recipes which can actually be executed` will be useful for EBF too.

draknyte1:
Huehuehue. 
I did this a year ago, because voiding output is stupid.

JohannesGaessler:
@SapientMC The recipe blocking in the EBF was caused by something else:
Before I fixed it, the EBF had only ever considered the first two ItemStacks in its input.

I did some more testing with recipe blocking, and it seems this fix is only necessary for recipes that have 2 or more fluid inputs.
So nothing apart from the Large Chemical Reactor needs to utilize it.

JohannesGaessler:
Only the last two commits are actually new.
I had merged unstable into this branch and when I then reworded one of the commits I made before that.

PR #1154, opened 2017-07-12
Distillery recipe fix

Antifluxfield:
Reduce the the batch size, to fix those recipes with more than 1000L fluid.

### DIFF
--- a/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java
+++ b/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java
@@ -2551,18 +2551,18 @@ if(Loader.isModLoaded("Railcraft")){
     	GT_Values.RA.addPyrolyseRecipe(Materials.Sugar.getDust(23), 								  GT_Values.NF, 					1, Materials.Charcoal.getDust(12),   Materials.Water.getFluid(1500), 				320, 64);
     	GT_Values.RA.addPyrolyseRecipe(Materials.Sugar.getDust(23), 								  Materials.Nitrogen.getGas(500),   2, Materials.Charcoal.getDust(12),   Materials.Water.getFluid(1500), 				160, 96);
     	
-    	GT_Values.RA.addUniversalDistillationRecipe(Materials.CharcoalByproducts.getGas(1000), 
-    			new FluidStack[]{Materials.WoodTar.getFluid(250), Materials.WoodVinegar.getFluid(500), Materials.WoodGas.getGas(250)}, 
-    			Materials.Charcoal.getDustSmall(1), 200, 64);
-    	GT_Values.RA.addUniversalDistillationRecipe(Materials.WoodGas.getGas(1000), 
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java
```diff
@@ -2551,18 +2551,18 @@ private void addRecipesApril2017ChemistryUpdate(){
...
-    			new FluidStack[]{Materials.CarbonDioxide.getGas(490), Materials.Ethylene.getGas(20), Materials.Methane.getGas(130), Materials.CarbonMonoxide.getGas(340), Materials.Hydrogen.getGas(20)}, 
-    			GT_Values.NI, 200, 64);
-    	GT_Values.RA.addUniversalDistillationRecipe(Materials.WoodVinegar.getFluid(1000), 
-    			new FluidStack[]{Materials.AceticAcid.getFluid(100), Materials.Water.getFluid(500), Materials.Ethanol.getFluid(10), Materials.Methanol.getFluid(300), Materials.Acetone.getFluid(50), Materials.MethylAcetate.getFluid(10)}, 
-    			GT_Values.NI, 200, 64);
-    	GT_Values.RA.addUniversalDistillationRecipe(Materials.WoodTar.getFluid(1000), 
-    			new FluidStack[]{Materials.Creosote.getFluid(500), Materials.Phenol.getFluid(75), Materials.Benzene.getFluid(350), Materials.Toluene.getFluid(75)}, 
-    			GT_Values.NI, 200, 64);
+    	GT_Values.RA.addUniversalDistillationRecipe(Materials.CharcoalByproducts.getGas(100), 
+    			new FluidStack[]{Materials.WoodTar.getFluid(250), Materials.WoodVinegar.getFluid(50), Materials.WoodGas.getGas(25)},
```
JohannesGaessler: I think the value for WoodTar should be 25.

JohannesGaessler (commented):
I think there has been a misunderstanding.
What I meant to be the solution was not changing the recipes themselves, but addUniversalDistillationRecipe:

if (maxFluidAmount > 1000) {
    fluidInputAmount = fluidInputAmount * 1000 / maxFluidAmount;
    fluidOutputAmount = fluidOutputAmount * 1000 / maxFluidAmount;
    solidInputAmount = solidInputAmount * 1000 / maxFluidAmount;
}

For example if you had a recipe with 3 Tiny Piles of Carbon Dust, 1500 Fluid input, and 2000 Fluid output:
The recipe quantity needs to then be reduced so that the maximum fluid quantity is 1000.
So the new quantities would be 1 Tiny Pile of Carbon Dust, 750 Fluid input, 1000 Fluid output.

Anyways, reducing the batch size would probably make the recipes more ergonomic and since the Carbon output is not that valuable, this change should be fine.

Antifluxfield:
@JohannesGaessler 
Well, I changed the recipes manually only because I'm confusing about how to get a proper solid output... 
For example, if we're going to half the size of a recipe with 1 small pile of carbon, do we have any universal method, with a result changing it to 1 tiny pile (which should be the most proper amount) instead of  simply dividing it by 2 and having no solid output?

PR #1156, opened 2017-07-12
Another Distillery recipe fix

Antifluxfield:
Only reduce the batch size if fluid amount is exceeding.
And this time I don't reduce the size of the normal recipes, because I found that we should reserve some space for overclocking.
@JohannesGaessler What about this one?

### DIFF
--- a/src/main/java/gregtech/common/GT_RecipeAdder.java
+++ b/src/main/java/gregtech/common/GT_RecipeAdder.java
@@ -9,6 +9,7 @@ import gregtech.api.enums.Materials;
 import gregtech.api.enums.OrePrefixes;
 import gregtech.api.interfaces.internal.IGT_RecipeAdder;
 import gregtech.api.objects.GT_FluidStack;
+import gregtech.api.objects.ItemData;
 import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_OreDictUnificator;
 import gregtech.api.util.GT_Recipe;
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/GT_RecipeAdder.java
```diff
@@ -562,6 +562,31 @@ public boolean addDistilleryRecipe(ItemStack aCircuit, FluidStack aInput, FluidS
...
+        			break;
+        		}
+        	}
+        	for (int i = tScale; i <= 10; i++) {
+        		if (aInput.amount % i == 0 && aDuration % i == 0 && aOutput.amount % i == 0) {
+        			tScale = i;
+        			break;
+        		}
+        	}
+        	aInput = new FluidStack(aInput.getFluid(), (aInput.amount + tScale - 1) / tScale);
```
JohannesGaessler, line 583: Why do you have this modifier for the input:
(aInput.amount + tScale - 1)

Antifluxfield: That's a ceil() after division. The input should not less than the exact amout of (amount / tScale).

draknyte1 (commented):
You just need a function to find th smallest denominators of all factors. 
Ie: 16:4:4 becomes 4:1:1.

Antifluxfield:
@draknyte1 
Not all recipes have such a good ratio: https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java#L3005
Recipes like this can't be lossless. So I only test some small factors, trying to make it lossless; if that fail, just let it be.

PR #1161, opened 2017-07-16
Diamond to Carbon rebalance

JohannesGaessler:
At some point during the first petrochemistry PR I have assigned a burn value to several materials, including Carbon.
This recently caused a disagreement on Discord about the realism of this behavior, and more pertinent to this PR, whether this behavior can be exploited.
Particularly, since 1 Diamond Dust can be electrolyzed into 64 Carbon Dust, the addition of a burn value for Carbon turned Diamonds into a very powerful fuel;
Electrolyzing and then burning one Diamond yields 1,536,000 EU at 100% efficiency.

By putting an Ore Drilling Plant on top of a Graphite vein it is currently possible to produce ~2930 EU/t, about as much as with 4 Pyrolyze Ovens.
For the calculation I made the following assumptions:
-The Graphite vein contains 80% Graphite, 10% Diamond, 10% Coal
-The Ore Drilling Plant only mines blocks from the Graphite vein
-A Sifter is used to increase output from Diamond Ore (+30%)
-Diamond Combs are not being used to increase output from Diamond Ore (+300%)

Two solutions were proposed to deal with this:
a) Remove the burn value for Carbon again
b) Reduce the conversion rate of Diamonds and Graphite into Carbon

Solution a) has been supported by Alkalus and Archengius, solution b) has been supported by me.
This PR is an implementation of solution b).

The argument for solution a) is that it is simpler.
The burn value on Carbon does not actually add anything to the game, since the materials that it is supposed to be made from, like Charcoal, can already be burnt themselves.
Therefore it is a useless feature and should just be removed again, which would be much simpler than having to deal with potential exploits.

The reason why I still support solution b) is because I think solution a) would not fix the fundamental problem, only a manifestation of it.
The fundamental problem is, as I see it, that the electrolyzation of Diamond Dust is simply too good as a source of Carbon.
Then, as soon as Carbon could be used in the same way as Charcoal, having large quantitites of Carbon had suddenly become useful.
Therefore, if the conversion rate of Diamonds into Carbon is not reduced, the same problem will reappear the moment having large amounts of Carbon becomes useful again.

Since I want to at some point introduce carbothermic reduction (removing oxygen from ores via carbon), I would like for the conversion rate of Diamonds into Carbon to be reduced regardless of whether or not the burn value for Carbon is kept.

What this PR would specifically change:
-Electrolyzing one Diamond Dust would yield 32 Carbon, down from 64
-Electrolyzing one Graphite Dust would yield 2 Carbon, down from 4
-Coal would be moved from the Electrolyzer to the Centrifuge. It now yields 0.5 Carbon and 94L of Natural Gas per Dust (total energy content unchanged). During the discussion the fact that Coal is a mixture of Carbon and Hydrocarbons had come up.
-Charcoal would be moved from the Electrolyzer to the Centrifuge. The output would remain unchanged (1 Carbon Dust per 1 Charcoal Dust)

As a side note, during the discussion on Discord the conservation of material had come up.
It was argued that since 64 Coal can be turned into 1 Diamond, the reverse process should yield 64 Carbon from 1 Diamond.
However, 1 unit of Coal is currently equal to 2 units of Carbon, so the amount of Carbon is inconsistent anyways.
With the recipe changes I propose one Diamond would consistently equal 32 Carbon.


Antillar:
Note that Carbon is not only obtained from diamonds, but also from Ashes and Dark Ashes, which are the leftovers in some recipes. These items are meant to be smth that did not burn during the processing operation. It is nonsence that something which is able to burn can be extracted from something which is unable to burn. Therefore I suggest removing fuel value from carbon and leaving the diamond to carbon recipes and diamond ore processing recipes as they are.

JohannesGaessler:
I agree that you should not be getting something that burns from Ash.
I have already changed the Ash recipes to no longer yield Carbon, see https://github.com/Blood-Asp/GT5-Unofficial/pull/1159

draknyte1:
You're sidestepping the actual issue at hand.

draknyte1 (changes requested):
This breaks carbon when your config option isn't even enabled. So no.

Dragon2488 (changes requested):
Read draknyte review above.

PR #1163, opened 2017-07-17
Drillers upgrade

Antifluxfield:
I try to make something that should be useful, including:
1) Oil Drill II & III, working on 3 * 3 and 6 * 6 chunks, matching the layout of the oil fields.
For all Oil Drills, you can set a minimum oil value by using Data Access Hatch and Programmed Circuits. Then the Oil Drills will stop working when the local oil field is near exhausted.
2) Concrete Backfiller and its advanced version, aiming for filling all underground caves. They should be helpful for reducing the rendering pressure and building a mob farm.
You can also adjust the radius of the advanced one, by putting Programmed Circuits into Data Access Hatch.


src/main/java/gregtech/api/enums/ItemList.java
```diff
@@ -638,7 +638,7 @@
     Casing_Coil_Cupronickel, Casing_Coil_Kanthal, Casing_Coil_Nichrome, Casing_Coil_TungstenSteel, Casing_Coil_HSSG, Casing_Coil_Naquadah, Casing_Coil_NaquadahAlloy,
     MobRep_LV, MobRep_MV, MobRep_HV, MobRep_EV, MobRep_IV, MobRep_LuV, MobRep_ZPM, MobRep_UV, Cover_PlayerDetector, Machine_Multi_HeatExchanger,
     Block_BronzePlate, Block_IridiumTungstensteel, Block_Plascrete, Block_TungstenSteelReinforced,
-    Honeycomb, Charcoal_Pile, Block_BrittleCharcoal, Seismic_Prospector, Seismic_Prospector_Adv, OilDrill1, OilDrill2, OilDrill3, OreDrill1, OreDrill2, OreDrill3, OreDrill4, PyrolyseOven, OilCracker, Crop_Drop_UUMBerry, Crop_Drop_UUABerry, Empty_Board_Basic, Empty_Board_Elite,
+    Honeycomb, Charcoal_Pile, Block_BrittleCharcoal, Seismic_Prospector, Seismic_Prospector_Adv, OilDrill1, OilDrill2, OilDrill3, ConcreteBackfiller1, ConcreteBackfiller3, OreDrill1, OreDrill2, OreDrill3, OreDrill4, PyrolyseOven, OilCracker, Crop_Drop_UUMBerry, Crop_Drop_UUABerry, Empty_Board_Basic, Empty_Board_Elite,
```
draknyte1: Why is it 1 & 3?

JohannesGaessler (changes requested):
If I interpret your code correctly, the advanced versions of the Oil drilling Rig pump Oil 9x/36x as fast as the base machine.
I feel like that will escalate pretty quickly.

JohannesGaessler:
@draknyte1 You can add `?w=1` to the github URL to ignore whitespace changes.

draknyte1:
> @draknyte1 You can add ?w=1 to the github URL to ignore whitespace changes.


@JohannesGaessler  Uhh, not sure what you're replying to?

JohannesGaessler:
Ah, sorry.
For me it showed ItemList as a +1500-1500 diff and all I saw from your review was `src/main/java/gregtech/api/enums/ItemList.java` so I assumed that was what you were pointing out.

Antifluxfield:
@JohannesGaessler
Not 9x/36x exactly... But still seems imbalanced.
Will rebalance.

By the way, those thousands of lines of diff scared me a lot... Its nice of you for that tip.

JohannesGaessler:
@Antifluxfield If you're using Eclipse you can also have it ignore whitespace changes under Window -> Preferences -> General -> Compare/Patch

Dragon2488:
why advanced version? Just make it behave like blast furnace, each voltage tier multiplies machine speed twice.

Antifluxfield:
@JohannesGaessler Thx. It works. And finished rebalance.

Dream-Master:
Any news here?

draknyte1 (changes requested):
I still don't think this should go into GT_MetaTileEntity_MultiBlockBase.java
public ArrayList<GT_MetaTileEntity_Hatch_DataAccess> mDataAccessHatches = new ArrayList<GT_MetaTileEntity_Hatch_DataAccess>();

It was added to the assembly line for a reason and if it's only used in another single machine, just add it there to the driller base class.

Antifluxfield:
So I make it like this. I comment out those codes in the MultiBlockBase, and move them to AssLine and DrillerBase respectively.
If anyone come up with some further usage of this hatch for most of the multiblocks in the future, then re-enable those codes; otherwise, just copy them for those multiblocks who need them.
@draknyte1

Blood-Asp:
@JohannesGaessler @draknyte1 Are the changes requested at your reviews done? As i can see it yes, but would prefer you to look too.

Blood-Asp:
And @Antifluxfield did you read #1189 ?

Antifluxfield:
@Blood-Asp 
Already fixed. 
Though I found another bug and fixed it in the latest commit...

JohannesGaessler (approved):

JohannesGaessler:
Right now I don't have the time to do an in-depth review, but what I talked about earlier seems to have been resolved.

Technus:
Well when using my code could you at least include me.

the undergroundOil(...) method useage case was taken from a scanner , which is meant only to give arbitrary information about the vein quality. (extractionSpeedCoef < 0 -> use only to read)

if you want to take fluids from veins -> use positive number that is a speed coefficient of how fast it should operate

that way you have automatic handling of configured values for each vein type.

```java
//Returns whole content for information purposes -> when drainSpeedCoeff < 0
//Else returns extracted fluidStack if amount > 0, or null otherwise
public static FluidStack undergroundOil(Chunk chunk, float drainSpeedCoefficient)
public static FluidStack undergroundOil(IGregTechTileEntity te, float drainSpeedCoefficient)
```

Technus:
Also after merging this,( or something else that Dream did last time) the underground fluids generation went borken values that i register dont match at all.

Techlone:
>extractionSpeedCoef < 0 -> use only to read

It's so wrong. Why did you this that? Why didn't you write two methods `getUndergroundFluid(ChunkPos)` and `decreaseUndergroundFluid(ChunkPos, int amount)`?
Also a changing the flow of fluid per an energy hatch tier is very strange. How does the tier affect on total amount of underground fluid?
>Also after merging this,( or something else that Dream did last time) the underground fluids generation went borken values that i register dont match at all.

Ha-ha. No, after [this](https://github.com/Blood-Asp/GT5-Unofficial/commit/82458fed2e4c065091f38340ceb05e98f28fc459) rebase by @Blood-Asp of yours commit underground fluids fucked up.
Everything was cool but no. Let's change it! To break existing worlds. Why?
image

Technus:
As you might noticed i removed the previous comment, since I was too tired to notice it was not misused but yes, another method wouldn't be bad.

The merging of my code actually fixed the previous UO/Pollution code, yes at the cause of the world breaking.  IIRC the previous code was doing things like regenerating fluid after draining the chunk to 0 and tikcing pollution even on unloaded worlds. It might've been done better but no one else stepped in to help at this time.

https://github.com/Blood-Asp/GT5-Unofficial/pull/1214 - new pullreq

PR #1168, opened 2017-07-22
Correct description of regulator

Antifluxfield:
Correct description of regulator.
The cost is actually tripled at:
https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/common/tileentities/automation/GT_MetaTileEntity_Regulator.java#L95

### DIFF
--- a/src/main/java/gregtech/common/tileentities/automation/GT_MetaTileEntity_Regulator.java
+++ b/src/main/java/gregtech/common/tileentities/automation/GT_MetaTileEntity_Regulator.java
@@ -26,7 +26,7 @@ public class GT_MetaTileEntity_Regulator
         		"Filters up to 9 different Items",
         		"Allows Item-specific output stack size",
         		"Allows Item-specific output slot",
-        		"Consumes 1EU per moved Item"});
+        		"Consumes 3EU per moved Item"});
     }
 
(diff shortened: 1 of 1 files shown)


Antifluxfield:
Rebalance the cumene recipe.

draknyte1:
Please try keep each change to seperate pull requests. Makes revision/reviewing easier.

Antifluxfield:
OK...

draknyte1 (changes requested):
So, is it meant to use 1eu or 3eu?
Might be better to fix EU usage rather than the description. Just depends what is actually intended I guess.

Antifluxfield:
@draknyte1
Well, I have reviewed the code of GT5.07.07, it shows that the regulator consumes 3EU per item since that time. Besides, the regulator is more powerful and more convienient than any other automation devices, so it's reasonable for it to consume more energy. Thus I think the 3EU-cost should be intended, so I changed the description instead of the EU usage.

SapientMC:
@Antifluxfield Could you please also improve Pump (block) tooltips? I think info about its pump area should be added. Numbers are here https://ftb.gamepedia.com/Pump_(GregTech_5_Unofficial)

mura3277:
@SapientMC Added by 9fc5d17

draknyte1 (approved):

PR #1170, opened 2017-07-24
Pyrometallurgy update

JohannesGaessler:
I have implemented new methods for processing ore that revolve around Blast Furnaces.
The general trend of these new methods is that they require more infrastructure and resources but in turn allow for a higher yield.
Generally for any ore that only partially consists of metals (Banded Iron Ore, Chalcopyrite Ore, etc.) it is now possible to invest more resources and in turn get as much output as from the pure ores (Iron Ore, Copper Ore, etc.).

To make this work I have overhauled Direct Smelting.
As before, all Ore Blocks yield 1 Ingot worth of material.
Crushed Ores and Dusts now consistently yield 6 nuggets worth of material.
This means that Impure/Purified Dusts no longer yield a full ingot, which I would consider an exploit since the regular Dust generally yielded less than that.
I have also added Direct Smelting for Magnetite -> Iron and Galena -> Lead.

~~I changed the chemical formula for Galena: Pb3Ag3S2 -> PbS~~

The new recipes generally work as follows:
Sulfuric Ores are processed in the Electric Blast Furnace with Oxygen to burn away the Sulfur and produce Oxidic Ores and Sulfur Dioxide (Roasting).
Oxidic Ores are processed in the EBF with Carbon to produce metal Ingots and Carbon Dioxide (Carbothermic Reduction).
Sulfur Dioxide and Carbon Dioxide can only be partially recovered; The amount depends on the tier of the Muffler Hatch.
Ores can also be processed in a Bronze/Bricked Blast Furnace where Sulfuric Ores as well as Oxidic Ores are converted into metal ingots in a single step.

Compared to the BBF the EBF:
-Processes Ores 10x as fast (6s instead of 60s per recipe)
-Yields CO2/SO2 as a byproduct
-Consumes only half as much Coal/Charcoal for processing ores (2 Small Piles of Carbon Dust instead of 1 Coal/Charcoal)
-Produces at least 5 times less Pollution
-Allows for further processing with Sulfuric Acid (Leaching, not yet implemented)

My changes would have the following consequences in terms of balance when using vanilla Smelting:

**Iron** would overall be easier to obtain. Magnetite, which is found very frequently, can now be easily processed into Iron via Direct Smelting. Banded Iron, Brown Limonite, and Yellow Limonite now yield less Iron when smelted in a vanilla Furnace (12 Iron Nuggets instead per Ore Block, down from 18 Iron Nuggets per Ore Block).

~~**Copper** would be slightly harder to obtain. Tetrahedrite, Chalcopyrite, and Malachite now yield 10 Copper Nuggets per Ore Block, down from 12 Copper Nuggets per Ore Block.~~
**Copper** would be unaffected due to the increase in Direct Smelting yield.

**Nickel** would be slightly harder to obtain. Garnierite and Pentlandite now yield 12 Nickel Nuggets per Ore Block, down from 12/18 Nickel Nuggets per Ore Block. I think this nerf would have little effect since it will only make a difference when you find a Nickel Vein at which point you will have more than enough Nickel anyways.

**Lead** would be slightly easier to obtain. Galena now yields 12 Lead Nuggets pre Ore Block, up from 8 Lead Nuggets per Ore Block.

~~**Silver** would be harder to obtain. When considering byproducts and not using Mercury the total amount of Silver per Galena Ore Block has decreased from 10 Nuggets to 2 Nuggets. When using Mercury the amount of Silver per Galena Ore Block decreases from 22.6 Nuggets to 14.6 Nuggets. When using Mercury the amount of Silver per Ore Block decreases from 22.6 Nuggets to 19.8. I think this nerf would have little effect since it will only make a difference when you find a Galena Vein at which point you will have more than enough Silver anyways.~~


Dragon2488:
Silver is component of electrum - perfect alloy for item transport pipes. Taking in account that there aren't much pure silver ore in galena veins, and 2 item pipes are made from 6 metal plates, item pipes are more expensive than now. I think they are already pretty costy, so silver processing should not be nerfed.

Dragon2488 (changes requested):
Silver is component of electrum, which is perfect alloy for item transport pipes. Pure silver ore in galena veins is not very common, and 2 item pipes are made from 6 metal plates, item pipes are more expensive than now. I think they are already pretty costy, so silver processing should not be nerfed.

SapientMC:
@Dragon2488 gold are much more rare than silver, so it will not become a bottle neck for electrum.

Dragon2488:
@SapientMC gold can be obtained in huge quantities from washing different ores with mercury, yielding 1 gold dust from one crushed ore with 60% chance. Galena electrolyzing gives silver in huge quantities too, just like gold from ore washing. When it will be disabled, there won't be any way to massively obtain silver.

JohannesGaessler:
@Dragon2488 You can obtain Silver from washing ores with Mercury, too.
Applicable Ores: Galena, Lead, Silver, Tungstate
About ~40% of the Silver in Galena currently comes from electrolyzing it.

src/main/java/gregtech/api/enums/Materials.java
```diff
@@ -538,6 +538,13 @@
...
+    public static Materials ArsenicTrioxide = new MaterialBuilder(619, TextureSet.SET_SHINY, "Arsenic Trioxide").addDustItems().setRGB(255, 255, 255).setColor(Dyes.dyeGreen).setMaterialList(new MaterialStack(Arsenic, 2), new MaterialStack(Oxygen, 3)).addElectrolyzerRecipe().constructMaterial();
+    public static Materials CobaltOxide = new MaterialBuilder(620, TextureSet.SET_DULL, "Cobalt Oxide").addDustItems().setRGB(102, 128, 0).setColor(Dyes.dyeGreen).setMaterialList(new MaterialStack(Cobalt, 1), new MaterialStack(Oxygen, 1)).addElectrolyzerRecipe().constructMaterial();
+    public static Materials Zincite = new MaterialBuilder(621, TextureSet.SET_DULL, "Zincite").addDustItems().setRGB(255, 255, 245).setColor(Dyes.dyeWhite).setMaterialList(new MaterialStack(Zinc, 1), new MaterialStack(Oxygen, 1)).addElectrolyzerRecipe().constructMaterial();
+    public static Materials AntimonyTrioxide = new MaterialBuilder(622, TextureSet.SET_DULL, "Antimony Trioxide").addDustItems().setRGB(230, 230, 240).setColor(Dyes.dyeWhite).setMaterialList(new MaterialStack(Antimony, 2), new MaterialStack(Oxygen, 3)).addElectrolyzerRecipe().constructMaterial();
```
draknyte1: I don't like this PR at all. 
You've overhauled the wrong way and only made Shit more realistic at the loss of balance and gameplay enjoyability.

draknyte1 (changes requested):

src/main/java/gregtech/api/enums/Materials.java
JohannesGaessler: >at the loss of balance and gameplay enjoyability.

Could you be a little more specific with what you mean?
These materials that you have marked only become relevant if you decide to process ores in an Electric Blast Furnace.
It is meant as optional way of obtaining more resources per Ore Block.

Unless I'm missing something the only tangible difference to the existing ore processing methods lie in the change to the Direct Smelting of Impure/Purified Dusts and the nerf to Galena.

MineAnPlay:
The impure and pure dusts gave more metal but less of other byproducts. Also silver is needed in large amounts. Yes I know about the mercury in bath recipes. I use it myself though you need a LOT more Silver then Lead. Yes I know it's not as realistic but PLEASE don't do realism just for realisms sake. This is a game it should be balanced around that. Right now a HV Motor requires 6 Silver but ZERO Lead. That's just one recipe. Lead is used for batteries and some recipes I think from IC2 (I'm in GTNH so not sure of vanilla recipes). IMO the PBF and EBF should be to BOOST the outputs though requires extra resources like carbon. Right now you can use calcite and brown limonite to get 3 iron per ore. It would be nice if more recipes alike was made.

JohannesGaessler:
@MineAnPlay 
When I made the evaluation for nerfing Galena I judged it to not have much of an impact because you don't need Silver for much.
In the default GT5U recipes you only need it in large amounts to produce Solar Boilers, Electrum Item Pipes, or Multilayer Fiberglass Circuit Boards.
As with everything, this PR is a suggestion for a change and I was aware that it would be controversial;
If the overwhelming majority of players opposes the change to Galena I have no problem with reverting it.

Also the Blast Furnace recipes do increase the output of ores, at least when you take the smelting of regular Dust as a baseline.

joakibj:
A note on side effects for mod compatibility. The PFAAGeologica mod does not generate any of the pure metal ores except gold. It only generates the sulfides, oxides, silicates and carbonates. With this change, early game will be a little harder, but not too big a deal if one can bridge this gap with a little infrastructure at the midgame. The whole idea of fun with GT, right?

However, removing the silver component from Galena would be quite harsh for PFAA players. As one requires a decent amount silver at MV - used for advanced circuits and mass producing basic circuits. Also later a preferred material for HV cables.

JohannesGaessler:
How about this:
Currently the chance for getting 1 Silver Dust from washing Galena is at 70%.
If the chance was increased to 100% you would get another 0.6 Silver per Ore Block.
This would increase the total amount of Silver per Ore Block to 2.22 (live: 2.37).

Dream-Master:
I think this is a good change and you get tons of quicksilver at this time all ready.

JohannesGaessler:
Perhaps we should have a discussion about where the baseline of Vanilla Smelting should lie.
When I decided on the numbers for this PR I had the following thought process:
-The pure Ores yield 2 Ingots per Ore Block
-The mixed Ores are supposed to be worse versions of the pure Ores
-Therefore the mixed Ores should yield less (2/3 of the pure Ores)

In my judgement this would give you enough resources from mining, but perhaps my sensibilities are simply vastly different from the rest of the playerbase.

I would like to hold an opinion poll:
Give this post a thumbs up if you think 2/3 of the pure Ore is a reasonable amount;
Give this post a thumbs down if you think 2/3 of the pure Ore is not enough.

Disclaimer: I do not have the authority to hold binding polls. The final decision will be up to Blood Asp.

MauveCloud:
@joakibj as far as obtaining silver when using PFAA, what about Garnet Sand?  Are you remembering to take that into consideration?

MineAnPlay:
Like I said I don't think anything should be changed from what it already is. The PBF and the EBF should be to BOOST the output. Like how using the EBF to make Iron using Brown Limonite and Calcite gives you 3 Ingots instead of 2 or putting Galena in the EBF so you get 32 nuggets of Silver and Lead with 16B of Oxygen per 8 Galena dust instead of 27 nuggets of Silver and Lead. Also another thing to keep in mind you can't even make Steam Machines with this update with out PURE Copper Ore since the others will give you nuggets. So I don't agree about the changes of the outputs of ores using the furnace. Right now you are just making everything more repetitive then making it more enjoyable.

Dream-Master:
@MineAnPlay run GTNH 1.5.0.7 test it (on Zeta or SP) and we can adjust stuff a bit if necessary.

JohannesGaessler:
>Also another thing to keep in mind you can't even make Steam Machines with this update with out PURE Copper Ore since the others will give you nuggets.

I think there has been a misunderstanding.
The Ore Blocks still give you ingots like they did before if you smelt them.
Also, my change will only start to make a difference once you start macerating your ores at which point you will by definition already have Steam machines.
(Outside of GTNH you can also craft nuggets into ingots)

leagris:
[Alkalus Edit - This is based on .27 information and isn't really relevant anymore]

Dudes, play survival from start!!!
See how much copper you need for the PBF

If your only source of copper is full pure ore,
3 copper ingots for 4 bronze
2 bronze ingots per plate
6 plates per bronze plated bricks
32 plated bricks
+ 12 plates for PBF controller
that's 204 bronze plates
408 bronze ingots
306 copper ingots

Now go find 306 pure copper ore ingots with basic pickaxes as your mining tool.

JohannesGaessler:
I've added a config for the resource level of mixed Ores.
They either yield as much as the pure Ores or 2/3 of the pure ores.
Using a Blast Furnace increases the yield by 50% for both cases.
The default for this config is for mixed Ores to yield as much as pure Ores.

For example, 1 Chalcopyrite Dust will by default smelt into 1 Copper ingot in the Vanilla Furnace.
In the Blast Furnace 2 Chalcopyrite Dust will smelt into 3 Copper ingots.

With the config enabled, 1 Chalcopyrite Dust will smelt into 6 Nuggets in the Vanilla Furnace.
In the Blast Furnace 1 Chalcopyrite Dust will smelt into 1 Copper ingot.

I also implemented the increased Silver yield from washing Ore with Mercury.
Anyone who needs large amounts of Silver will already be washing their Galena Ore with Mercury and will overall get the same amount of Silver as before.

Dragon2488 (approved):

PR #1171, opened 2017-07-24
Pyrolyse oven coil scaling

JohannesGaessler:
I think currently the Pyrolyse Oven becomes available too late in the game because it requires Kanthal Heating Coils.
This change would make it so that the Pyrolyse Oven can use any Coil tier.
Therefore it would become available at MV.
I made it so that the processing speed of the Pyrolyse Oven scales linearly with Coil tier:
CuNi: 50%, FeAlCr: 100%, Ni4Cr: 150%, Fe50CW: 200%, etc.
EU/t is not changed; With CuNi the energy cost of making the charcoal is at ~20% of the enrgy gain.

Personally I think that the Coil tier with 100% could even be CuNi;
These values are meant as a compromise for Blood Asp since he seemed to oppose making the Pyrolyse Oven available earlier.
A Pyrolyse Oven at 100% speed produces 750EU/t at 100% efficiency.

I reduced the Pollution per Second for one Pyrolyse Oven from 400 to 200 to make building more than one more viable.

### DIFF
--- a/src/main/java/gregtech/common/GT_Pollution.java
+++ b/src/main/java/gregtech/common/GT_Pollution.java
@@ -50,7 +50,7 @@ public class GT_Pollution {
 	 * Large Boiler(240)
 	 * Large Gas Turbine(160)
 	 * Multi Smelter(100)
-	 * Pyrolyse Oven(400)
+	 * Pyrolyse Oven(200)
 	 * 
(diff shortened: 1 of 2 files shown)


Dragon2488 (approved):

draknyte1 (approved):
I like how you fixed the German spelling errors.

PR #1172, opened 2017-07-24
More Item Pipe Materials

JohannesGaessler:
I think currently the materials from which you can make Item Pipes are unneccessarily restrictive.
This PR would enable making Item Pipes from the following materials:
-Tin with a base Item Capacity of 1
-Nickel with a base Item Capacity of 1
-Cobalt with a base Item Capacity of 2
-Aluminium with a base Item Capacity of 2


Dragon2488:
Pipes from Tin and Aluminium sound strange.

JohannesGaessler:
I don't think it's any more strange than making a pipe from Electrum.

Dream-Master:
I using Tin Pipes in my GTNH Modpack for ages so add more various pipes why not. Gt++ add more pipes too.

leagris:
Le 25/07/2017 à 11:14, Martin Robertz écrivait :
> I using Tin Pipes in my GTNH Modpack for ages so add more various pipes
> why not. Gt++ add more pipes too.

And instead of trying to expand GT into reality grinding you actually
reworked the piping system so it compete with other mods decently?

Fluid pipes are highly broken in their working.

Why not merge item and fluid pipes and energy into a single conduit tile?

Why not use covers on conduit to determine what interface it has with
connected machine?

Ductible/Elastic ring: Fluid interface
- Wood / Rubber / Polysiloxane / Polybutadiene ring conver = connect fluid

No interface for Items

Power plug as interface for Power

Then decide on capabilities for your conduit. Like machine block can
transport everything.

Keep pumps and conveyors as active interfaces.

Add plugs capability to use IC2 EU. Then you have a solution to the
active polling needed with IC2 EU. No need to get all the cable be
compatible. Just the power plug cover needs to.

Then rework fluid transport to be efficient. Allow different kind of
distribution,

Keep current pipes for backward compatibility. Add whole new conduits
instead.

Oh, and allow your conduits to be painted any texture with an IC2
obscurator or a printer.

To sum it up:

- Stop adding more grind.

- Add content

- Fix broken features

-- 
Léa Gris

draknyte1:
I think the choice of materials is strange, Let me think of other possible choices.
Are any of these in use:

> Cupronickel
> Brass
> Thaumium
> Wrought Iron

Otherwise I think this is okay.

draknyte1 (approved):
Consider the other material options, especially Wrought Iron.

PR #1182, opened 2017-08-01
Circuit/Chemistry Update follow-up

JohannesGaessler:
This PR represents a follow-up to the Circuit changes done by Blood Asp and the Chemistry changes done by me.
It has unfortunately become bigger than I originally intended.

This PR has the following goals:
1. Make Circuits cheaper
2. Make Circuit board costs progessive
3. Give the player a reward for building the MV Circuit Assembler beyond the ability to progress
4. Implement a recipe for the Wetware Circuit Board that does not need beer
5. Give earlier access to the Vacuum Freezer
6. Move content from the MV -> HV -> EV transition to the EV -> IV transition
7. Give the player easier access to Niobium
8. Make chemical processing lines more convenient
9. Make minor tweaks to early game circuits

# 1. Make Circuits cheaper
I think the cost for making circuits is currently too high, particularly for the HV -> EV transition where you currently need 5 Mainframes.
I have taken the following steps to make Circuits cheaper:
* Base Circuits from the Integrated Processor and onwards now only require 2 of each electronic component, down from 4.
* The Copper cost of Small Coils has been reduced to 50%, can be reduced to 25% by replacing the Steel with Nickel-Zinc Ferrite (new material).
* The cost of SMD Capacitors has been reduced to 25%.
* The conversion rates of base Circuits into higher tiers have been adjusted to make the conversion into an Assembly cheaper.
* Integrated Processors, Nanoprocessors, and Quantumprocessors now each use Circuit Boards that are cheaper by 1 tier. Integrated processors use Plastic Circuit Boards (new), which can be produced from either Polyethylene (100% cost), Polyvinyl Chloride (50% cost), or Polytetrafluoroethylene (25% cost).
* SoC recipes no longer use fine wires that are 1 tier more expensive than the regular recipe. Since all you save with a SoC are Resistors, Capacitors, and Transistors, I don't think it's worth using SoC recipes if the wiring becomes more expensive.
* Wetware Circuits now only need 1 Wetware Circuit board, down from 2

As an example for the cost reduction, let us look at the Integrated Processors:

| Material | Integrated Processor | Processor assembly | Workstation | Mainframe | EV Circ. Ass. + Vac. Fr. |
| --- | --- | --- | --- | --- | --- |
| Aluminium | 0.25 -> 0.03 | 0.75 -> 0.06 | 1.50 -> 0.19 | 9.50 -> 3.13 | 47.50 -> 6.82 |
| Copper | 0.84 -> 0.54 | 8.41 -> 3.22 | 17.44 -> 10.28 | 79.75 -> 48.13 | 398.75 -> 127.09 |
| Epoxy Resin / Polyethylene | 1.00 -> 1.00 | 4.00 -> 3.00 | 10.00 -> 11.0 | 40.00 -> 44.00 | 200.00 -> 121.00 |
| PVC / Silicone Rubber | 1.00 -> 0.13 | 3.00 -> 0.25 | 6.00 -> 0.75 | 30.00 -> 4.50 | 150.00 -> 11.25 |
| Redstone | 2.00 -> 1.00 | 12.00 -> 5.00 | 24.00 -> 15.00 | 96.00 -> 60.00 | 480.00 -> 165.00 |
| Silicon | 0.25 -> 0.25 | 1.00 -> 0.75 | 2.25 -> 2.50 | 10.00 -> 11.00 | 50.00 -> 29.50 |
| Soldering Alloy | 0.50 -> 0.50 | 2.50 -> 2.00 | 6.00 -> 7.00 | 26.00 -> 30.00 | 130.00 -> 81.00 |
| Total | 6.22 -> 3.64 | 34.84 -> 14.84 | 74.34 -> 49.22 | 322.00 -> 210.88 | 1610.00 -> 569.41 |
| Total cost reduction | 41.5% | 57.4% | 33.8% | 34.5% | 64.7% |

Table entries depict the cost in Dusts/Ingots.
The total cost calculation makes the estimation that every material has the same cost;
It just sums up the total amount of material required.
Carbon, Electrum, Gallium, and Steel costs are not listed separately because their contribution to the total cost is negligible;
They are still factored into the total cost figure.
Currently 5 Mainframes are required for the EV Circuit Assembler and the Vacuum Freezer;
This PR would make it so that you need 2 Mainframes and 3 Workstations.

As we can see, the cost reduction is most prevalent for the lower Circuit tiers.
The idea behind this is to attach a bigger reward to unlocking a new Circuit tier.
If the Base Circuits are cheaper, the cost of making the same Circuits will go down further with each new tier.

# 2. Make Circuit Board costs progressive
Currently the Fiberglass Circuit Board is much cheaper than the Epoxy Circuit Board.
I think it would be better if the cost of Circuit Boards were to steadily increase with each tier.
Also, because Epoxy Resin is only needed for Integrated Processors, the setup for producing it currently becomes rather useless once you unlock Nanoprocessors.
This PR would make it so that Fiberglass Circuit Boards are also made from Epoxy Resin, but with the addition of Carbon Fibers or Glass Fibers (new).

Glass Fibers are made from Borosilicate Glass (new), which in turn requires Boron (1 Boron per 64 Glass Fibers).
Because currently the only way of obtaining Boron is from pulverizing Lepidolite, I have added Borax as a byproduct to Salt and Rock Salt.

# 3. Rewarding the player for the MV Circuit Assembler
Considering the immense infrastructure cost of the MV Circuit Assembler (Cleanroom, etc.) I think the reward for completing the task is too low.
I think an appropriate reward would be a reduction in the production cost of Circuits.
Currently making Integrated Processors is rarely worth the effort because Integrated Logic Circuits are cheaper and require less infrastructure.
Also, there is no cost reduction for LV circuits since they cannot be crafted with the new infrastructure.
This PR would decrease the cost of Integrated Processors to make them more appealing (as detailed above).
It would also add a new LV Circuit: The Microprocessor.
Microprocessors are crafted very similarly to Integrated Processors, except they use Copper instead of Red Alloy.
Their cost is at ~50% of that of Integrated Logic Circuits.

# 4. Beer-free Wetware Circuit Board recipes
This PR would add recipes for the Wetware Circuit Boards that do not require beer.
Instead they require a mixture of Water, Sugar, Meat, and Salt.
Dragon Blood can still be used as an alternative if desired.

# 5. Earlier Access to the Vacuum Freezer
I think currently the Vacuum Freezer becomes available too late in the game.
You only gain access to it at the end of HV because you need 3 Mainframes to construct it.
This PR would make it so that the recipe instead uses 3 Workstations, which would make the Vacuum Freezer available as soon as a MV Circuit Assembler is constructed.
I think this change would give the player more flexibility for the order in which tasks are completed;
The player would have the option to construct the Vacuum Freezer earlier in the game to upgrade their Electric Blast Furnaces, Multi Smelters, or Pyrolyze Ovens earlier.
Also, I think earlier access to the Vacuum Freezer would decrease the amount of time that the player needs to wait for Kanthal, Nichrome, etc. because the player could start production earlier.
The following machines would as a consequence become available earlier:
* Diesel Engine
* Large Heat Exchanger
* Large Titanium Boiler
* Ore Drilling Plant

# 6. Content shift to EV -> IV
I think currently there is too much that you need to do for the MV -> HV -> EV transition, and too little that you need to do for the EV -> IV transition.
This PR would make it so that you only need to start producing Epoxy Resin at EV for the IV Circuit Assembler.
Also, the more complicated Cable insulation recipes would only start at IV Cables.

Recently it has been argued that the more complicated Cable recipes should be disabled by default.
I think they should be enabled by default because my stance is that all content should be enabled by default and that the player should then decide which pats to disable.

# 7. Easier Niobium Access
Currently Niobium can only be obtained as a byproduct from Tantalite, which in turn can only be found in Manganese veins.
This PR would make it so that Niobium can also be obtained from Pyrolusite, Apatite, and Phosphorous processing.
Also, the Phosphate Ore in Apatite veins would be replaced with Pyrochlore (new, contains Niobium).
Since Phosphate seems to be rather useless I don't think removing it from the mix would be a problem.

# 8. More convenient chemical processing lines
Currently setting up an automated processing line for recipes that work with Fluid Cells is rather inconvenient.
This is to a large part because you need a lot of Fluid Canners to fill or empty Fluid Cells so that they can be processed by the next machine.
This PR would make it so that Chemical Reactors gain the ability to automatically can fluids at the end of their recipe, controlled by an Integrated Circuit:
* A Circuit config of 1 executes the recipe normally
* A Circuit config of 11 cans the output Fluid into an Empty Fluid Cell. If the recipe puts out a Fluid and Fluid Cells, they are swapped.
* A Circuit config of 2 voids any Water output
* A Circuit config of 12 voids Water and cans the remaining output
I estimate that this change will cut the amount of Fluid Canners required by ~50%.

~~Because it would be annoying to have to put an Integrated Circuit into each machine, I have made the following change to how basic machines process their inputs:
If at least one of the machine's input slots are empty, and if the inputs do not already contain an Integrated Circuit, an Integrated Circuit with config 0 will be added to the input.
If two inputs contain the same Item it will be treated as one of the slots being empty.~~

~~This change would make it possible to differentiate between a recipe that uses two Items as an input and a recipe that only uses one of those items as input.
Also, it is now possible to give recipes a default behavior that can then be changed by adding an integrated circuit.~~

I also added more convenient recipes to produce instrumental materials like Acetone to make plastic production easier.

~~Edit:
Because it has been requested, here is a detailed explanation of the consequences that automatically an Integrated Circuit would have in regards to backwards compatibility.
For there to be any problem a recipe would have to malfunction in one of the following two ways:~~
~~1. A recipe does not work when it's supposed to.~~
~~2. A recipe is executed when it's not supposed to.~~

~~The first case will never happen because adding an additional item to the machine input will not impede recipes that could have been executed without the additional item.
The second case can also be expected to never happen because Integrated Circuits with a config of 0 are almost never used, due to the inability to set an Integrated Circuit to a config of 0 with a screwdriver.
The only instances I could spot is in the Distillery with the old Sulfuric Gas recipes (no conflict) and with dehydrating Lithium Hydroxide in GT++ (no conflict).
Also, even if there were two conflicting recipes literally all you would have to do to resolve it would be to put Cobblestone into the empty machine slots.
As a player I would prefer this over having to put Integrated Circuits into 10 other machines.~~

# 9. Early game circuits
I have made the following changes to the production of Electronic Circuits

-The recipe time for Glass Tubes has been reduced: 12s -> 8s
-With a Basic Assembling Machine there are cheaper recipes for Coated Circuit Boards (-62.5% Sticky Resin cost), Vacuum Tubes (-33% Copper/Paper cost), and Phenolic Circuit Boards (-20% Glue/Bisphenol A cost)

# 10. TL;DR
Circuits are cheaper, Plastic production is less complicated


MineAnPlay:
Can it be made so the Large Chemical Reactor can do multiple recipes? As in if you input the required fluids/items with a P.C. and it will do both recipes and output the finally product? Because right now making 10+ Large Chemical Reactors is kinda excessive.

JohannesGaessler:
The Large Chemical Reactor can in theory already do several recipes at once.
The only problem is that there is no convenient way to direct outputs to specific hatches/buses.
I plan to at some point add the ability to lock a Fluid to an Output Hatch.

draknyte1:
> Also, the Phosphate Ore in Apatite veins would be replaced with Pyrochlore (new, contains Niobium).
> Since Phosphate seems to be rather useless I don't think removing it from the mix would be a problem.


This will break oregen on existing worlds, which is generally considered a bad move.
Why not just make Pyrochlore a component of Phosphate when processed in a certain way?

> Because it would be annoying to have to put an Integrated Circuit into each machine, I have made the following change to how basic machines process their inputs:

Probably not a good idea to change how all machines process things..

JohannesGaessler:
>This will break oregen on existing worlds

How will this break oregen?
Is there any functional difference to changing the config entry?

>Probably not a good idea to change how all machines process things..

I think it is a good idea because currently it's impossible to specify recipes where one of the slots is left empty.

draknyte1:
For existing worlds it will require new chunks to generate. If worlds are pregenerated then it's unobtainable. 

What about recipes that use circuits?

MineAnPlay:
Also can you make the Large Chemical Reactor not void it's output when it's full? Because that is one issue that is quiet annoying to fix.
Also about the wetware change you should make that new item a bit more expensive since IMO Dragon Blood route should be better for those that do it that way.

leagris:
It also makes crafting Gregtech matches and matchbox impossible as these depends on phosphorus dust.

JohannesGaessler:
@draknyte1 
Pyrochlore is obtainable even in pregenerated worlds.
Perhaps my wording was chosen poorly;
Apatite and Phosphorous give you Pyrochlore as a byproduct and that's how you can obtain Niobium from those ores.
Also, the only use of Pyrochlore lies in being a source of Niobium, so I think the only relevant question is whether you will get enough Niobium in pregenerated worlds.
Since Apatite yields 8 Crushed Ore per Ore Block the total amount of Niobium per Apatite Ore is at ~0.16.
By comparision, 1 Pyrochlore Ore yields ~0.36 Niobium.

Regarding the integrated circuits, I think I might have come up with a better solution.
Instead of using an Integrated Circuit with config 0 I think it might be better to explicitly add an item that corresponds to an empty slot.
This item would be guaranteed to not show up accidentally in recipes.

Or perhaps I should drop the feature alltogether.
If you needed an Integrated Circuit for most of the Chemical Reactors that would mean that you require 3 Circuits per machine instead of 2.
But since I also made Circuits cheaper this might be okay.

@MineAnPlay 
>Also can you make the Large Chemical Reactor not void it's output when it's full?

I'm not 100% sure what you mean. 
I'm assuming you mean that you want the Large Chemical Reactor to no longer void its Fluid output if the first two Fluid Hatches already contain Fluids. 
This has already been fixed here: https://github.com/Blood-Asp/GT5-Unofficial/pull/1153 
Apparently Dreammaster has overlooked this PR.

>Also about the wetware change you should make that new item a bit more expensive since IMO Dragon Blood route should be better for those that do it that way.

Granted, for people that have already built large setups to get Dragon Blood this change might be upsetting because they'll effectively have wasted their time by comparision.
However, since the materials you need for the new recipe (Water, Sugar, Mince Meat, Salt) are all easily obtainable the only way to really make it more expensive would be to add another ingredient that is harder to obtain.
I think a better solution would be to increase the conversion rate of Dragon Blood into Sterilized Growth Medium, so the investment made into Dragon blood still pays off.

@leagris 
In Gregtech there are the following three materials:
* **Phosphor** [P], which corresponds to the real-life element named phosphorous
* **Phosphorous** [Ca3(PO4)2], which corresponds to the real-life material named tricalcium phosphate
* **Phosphate** [PO4], which corresponds to the real-life anion of phosphoric acid named phosphate.

Only Phosphor and Phosphorous can be used to make Matches.
Phosphate cannot be used.
In fact, the only use I could spot for Phosphate was making Fertilizer or as a source of Phosphor.
Since you get way more Phosphor than you could ever need from an Apatite vein I think changing the Phosphate to Pyrochlore would not take away anything from the player.

Techlone:
Why does this recipe exist?
image
It's useless. You can obtain more the fine wires than the 1x wires at any game stage. Maybe remove it?

Blood-Asp:
The earlygame fine wire recipe was added later.

src/main/java/gregtech/api/util/GT_Recipe.java
```diff
@@ -160,7 +160,7 @@ protected GT_Recipe(boolean aOptimize, ItemStack[] aInputs, ItemStack[] aOutputs
         mFluidInputs = aFluidInputs;
         mFluidOutputs = aFluidOutputs;
         mDuration = aDuration;
-        mSpecialValue = aSpecialValue;
+        mSpecialValue = aSpecialValue;
```
draknyte1, line 163: Probably just revert this class.

draknyte1 (changes requested):
Generally happy with this PR.
I still think `gregtechproxy.mEasierIVPlusCables = tMainConfig.get("general", "EasierIVPlusCables", false).getBoolean(false);` should be default false.

Muramasa-:
@JohannesGaessler This isn't a bug with your PR, but shouldn't this recipe use RedAlloy and not RedstoneAlloy? https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java#L467

Blood-Asp:
Afaik this PR is ready to merge. Will do that in a few hours. If anyone has a problem with that, tell me why asap.

PR #1184, opened 2017-08-02
Texture overhaul

Technus:
Adding paging of textures (of hatches) in the least breaking way i could come up with.

only thing to make it work is to make sure you set the mTexturePage at the same time as the mMachineBlock.

GT5u range uses page 0 to avoid me changing stuff.

I decided to use page 8 for my addon.

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch.java
+++ b/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch.java
@@ -1,5 +1,7 @@
 package gregtech.api.metatileentity.implementations;
 
+import gregtech.GT_Mod;
+import gregtech.api.enums.GT_Values;
 import gregtech.api.enums.Textures;
 import gregtech.api.interfaces.ITexture;
 import gregtech.api.interfaces.tileentity.IGregTechTileEntity;
@@ -66,7 +68,7 @@ public abstract class GT_MetaTileEntity_Hatch extends GT_MetaTileEntity_BasicTan
         actualTexture=aNBT.getByte("mMachineBlock");
         mTexturePage=aNBT.getByte("mTexturePage");
 
-        if(mTexturePage!=0 && getBaseMetaTileEntity().isServerSide())
+        if(mTexturePage!=0 && GT_Values.GT.isServerSide())
             actualTexture|=0x80;//<- lets just hope no one needs the correct value for that on server
         mMachineBlock=actualTexture;
     }
--- a/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch.java
+++ b/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch.java
@@ -6,8 +6,10 @@ import gregtech.api.interfaces.tileentity.IGregTechTileEntity;
 import net.minecraft.nbt.NBTTagCompound;
 
 public abstract class GT_MetaTileEntity_Hatch extends GT_MetaTileEntity_BasicTank {
+    @Deprecated //Use texture changing method
     public byte mMachineBlock = 0;
-    public byte mTexturePage = 0;
+    private byte mTexturePage = 0;
+    private byte actualTexture = 0;
 
     public GT_MetaTileEntity_Hatch(int aID, String aName, String aNameRegional, int aTier, int aInvSlotCount, String aDescription, ITexture... aTextures) {
         super(aID, aName, aNameRegional, aTier, aInvSlotCount, aDescription, aTextures);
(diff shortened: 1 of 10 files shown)


Technus:
with this you can use textures with id's from 0 to 15*128 
(can be extended by changing the array size of CASING_BLOCKS just like shown in the last commit)

the method abuses currently using channel for textures

If 8th bit of byte is set then the data is for page, else for texture index on that page.

Technus:
When using the multiblock base functions just give em this (thank god they accept integer...)
`(newPageID<<7) | newTextureID`

Or when manually editing mMachineBlock of hatch - edit the mTexturePage also so it stays correct.

Technus:
Worst thing that can happen:
- mTexturePage page did not get updated (since the java src doesnt know bout it)- 
which might lead to stuff like missing texture - which results in transparency. (this only happens when you attach the hatch that was using page !=0 to multiblock that uses page 0 but not actually sets it)

Technus:
@draknyte1 
@JohannesGaessler 
If someone else needs notice to update their coding style (i mean when you update the mMachineBlock directly instead of using the multiblockBase methods) attach them here @.

Technus:
-fixed using wrong field names in comments above-

draknyte1:
This isn't a proper fix for me, because I need backwards compat. 
It's easier to resolve than this if you're only targeting current version++.

Another option is to pass the texture array into the hatch functions and if null/not specified in secondary function, use default.

draknyte1 (commented):
I think that this is over-complicated for what you're trying to achieve.

Expand array to Short.MAX_VALUE, swap byte methods to short, leave old methods with byte to pass to short methods if you so desire. No texture page garbage this way and nothing breaks because everything downcasts to byte as it is so we get no breakage.

Technus:
This might cause issues when old code tries to read the short into some byte.

My solution is basically what you want to do but instead of changing to short i add new field.

The methods u speak of (i assume the ones like addmaintenancehatch fron multiblocl base) already use integer as input and that can be used to set the page and texture at once.

I think that if i would male the old casing texture array unchanged and made a new longer array it wpupd be even more compatible.

Still the short solition would be best if the field was designed as private since the beggining (so u could enforce method use).

One more thing to consider is the network packet size/compatibility.

I think best attempt would be to use the old field as a placeholder for compatibility. Creation of new short field that would be updated periodically if the old byte changed. And use of the new field as texture pointer. While to prevent issues the new field would have to be private to enforce use of a method that silently update the old byte state so it does not trigger the short update that would limit it to one byte.

Awaiting input.

draknyte1:
Again, this doesn't solve the issue for me and only makes for a headache in terms of compatibility. 

I'm aware ints get cast to byte and my suggestion is easier to implement and holds better compatibility. New fields are a pain in the ass to access because I'd always prefer avoiding reflection. 


RDIT: internal Dev. Team is discussing a solution and this will probably get closed in favour of something more flexible for backwards compatibility. Resolve pending.

JohannesGaessler (commented):
My intuitive solution to this problem would have been one of the following:
* Replace the byte for the machine block with a short
* Interpret the byte for the machine block as the unsigned value, extending the range to 255

The problem with both of these approaches lies with the expected value of mMachineBlock.
Currently it can be expected to always be >=0.
If you were to interpret the byte as unsigned or if you were to downcast a short/int the result could very well end up being <0.
Since Greg did not make use of getters/setters, mMachineBlock being <0 poses a major threat to backwards compatibility, especially since it is meant to be used as an array index.

I think organizing the textures in pages is a better solution after all.
However, there needs to be documentation about what pages are being used by whom to avoid conflicts.

Working with a private short to resolve the texture would only complicate things at this point, I believe.
Since you need to keep mMachineBlock as an access point for backwards compatibility you still would not have a well-defined interface for the texture ID.
All you're getting is more work because now you need to make sure that mMachineBlock and the new short are synchronized.

After looking at this PR in detail I also think that it would be possible to make adjustments so that GT++ works with this PR and also with any previous GT5 version.

I think the optimal solution would be to allocate the first half of page 0 and also the entirety of page 1 for GT5U.
The second half of page 0 would then be allocated for GT++, GT5U textures currently in that range would be moved to page 1.
Currently the only textures in that range are the Heating Coils and the textures for the Large Chemical Reactor from what I can tell.
The textures on page 0 can be set and used the same whether there actually are more pages or not since you would only be working with an index in the old [0,127] interval.

Technus:
The problem then might arise - when you have to revert the block state from page 1+ to page 0, but the code doesn't edit the page value.

But I have a slightly cheeky solution for that too. Look at last commit, it should solve most of the compatibility issues I hope.

src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch.java
```diff
@@ -79,6 +79,7 @@ public final void updateTexture(byte texturePage, byte machineBlock){
     @Override
     public final void onValueUpdate(byte aValue) {
         mMachineBlock=actualTexture=(byte)(aValue & 0x7F);
+        mTexturePage=0;
```
draknyte1: This isn't deprecated?

draknyte1:
When you're satisfied, built up a test jar for me (at work) and I'll see how we go.
Prior to your changes within the last day, It would simply crash outright.

src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch.java
Technus: in case someone uses onValueUpdate revert the page to 0
in the updated code -> the next line should change the page to x

similar logic applies to the checking of mMachineBlock field for changes - is it deprecated so turns page to 0

Technus:
Yes i got pm from dream about the crash

Technus:
I asked DreamMaster to check if it will work out of the box with GT++

draknyte1:
> I asked dream to check if it will work out of the box with GT++

Just build a jar for me and I'll tell you in five minutes flat.
I'm at work and can't actually pull/clone from github, it's blocked.

Dream-Master:
http://jenkins.usrv.de:8081/

Technus:
http://jenkins.usrv.de:8081/job/Gregtech-5-Unofficial/

Technus:
Latest ones.

draknyte1:
Those require DreamCraft and after making a deobf'ed version for my dev enviro forge fails to load it.
It continues asking me for any version of DreamCraft to be present...
@Dream-Master @Technus 

```
Caused by: java.lang.NoSuchFieldError: SENTIENT_ENDER
	at com.dreammaster.gthandler.GT_CustomLoader.fixEnderIO(GT_CustomLoader.java:194)
	at com.dreammaster.gthandler.GT_CustomLoader.run(GT_CustomLoader.java:169)
	at com.dreammaster.main.MainRegistry.PostLoad(MainRegistry.java:307)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```
I cannot even test this because DreamCraft crashes due to me having an older version of EIO.
Damn.

Side note, when did `GT_Utility.getBlockFromStack(this.getItem())` Get changed to accept an ItemStack? it's the only error I have when using GTNH on build path.

Technus:
I have to check that method, but it looks like it was not our change i see that present in main repo too. Will also try to get you a build of this branch when i get home.

Technus:
That branch in current state
[libs.zip](https://github.com/Blood-Asp/GT5-Unofficial/files/1202173/libs.zip)

Technus:
Just as a heads up i am using negative values of mMachineBlock - but only server side so it should not break things by trying to access textures with negative index (the network handler strips the sign bit)

draknyte1:
I am now using 96-127 and we're planning to migrate 64-95 so that I get that block of 64 total.
Since .08.33 and .07.07 have the last 64 free, this will only be an issue for any version of .09 I support prior to the change here. Although It's not a massive issue right now, as I only use 32 textures (But I am looking to expand to more already, just not in a hurry until we resolve textures), I am waiting for a final .09.31 release so that I can have the modpack I play finally updated from .27 (Only place the issue will exist but I can drop support for it next GT version.)

My array index was changed here - https://github.com/draknyte1/GTplusplus/commit/afc205f81eb3b6fa45d114e7a6439bed10c7da44#diff-70b7f313f85f0f4b252f7606af0dfac6

draknyte1 (approved):
I am happy with this assuming compatibility is now fixed.

PR #1193, opened 2017-08-17
Improve fluid regulator

Antifluxfield:
It's always annoying rightclicking hundreds of times when someone is trying to config a regulator with a high pump speed. So I make the cover able to increase/decrease 256 per click when sneak-rightclicking with screwdrivers.


Exidex:
Maybe add that info to the tooltip?

Antifluxfield:
Seems the metaitems don't support multiple lines of custom tooltip...

Dragon2488:
They do. Don't except text to magically break into lines, okay?
You have to manually split it with \n

Antifluxfield:
@Dragon2488 
image
Do you mean sth like this?
I've already tried so and failed, so I gave that conclusion. It's not a simple issue can be solved with an LF.

Exidex:
@Antifluxfield you could add separation logic to the GT_MetaBase_Item#addInformation and for example use `////` as separator.

PR #1195, opened 2017-08-18
Fix inconsistent worldgen

Techlone:
An ore vein may be generated in the center chunk when a neighbor chunk is generating and the random depends of the neighbor chunk coordinates but it should depend of the center chunk coordinates.


draknyte1:
The only issue I have with fixing oregen is it makes existing maps inconsistent. Otherwise nice catch.

Blood-Asp:
As the current worldgen is fully random, independed of the worldseed, there is nothing that this PR makes oregen more inconstistent...

PR #1197, opened 2017-08-19
Fluid Canner cost reduction

JohannesGaessler:
I think Fluid Canners are currently too expensive.
Typically they do not actually process anything, they are just there for utility or because you need a certain Fluid as a Cell or vice versa.
As such I think a cost of 2 Circuits and 2 Pumps is not adequate.
This PR reduces the cost to 1 Circuit and 1 Pump.

This is what the new crafting recipe would look like:


tomasbrod:
Thank you!

PR #1199, opened 2017-08-21
Multi-lines tooltip support

Antifluxfield:
Use "/n " to seperate the tooltip to multiple lines.
Thanks @Exidex for your hint.


Exidex:
I guess it makes more sense to use `\n` as separator

Antifluxfield:
@Exidex Not a good idea. It will break the localization.

Dragon2488:
/n will not break anything, but looks much more familiar

Antifluxfield:
@Dragon2488 Sounds reasonable.

PR #1209, opened 2017-09-10
Odp speed improving

Techlone:
I forgot to add a changing of a basic speed for each ODP.
Now Mk4 is 2 times faster than Mk1. Basic speeds (for LV):
Mk1 - 960 ticks
Mk2 - 800 ticks (-16%)
Mk3 - 640 ticks (-20%)
Mk4 - 480 ticks (-25%)


Antillar:
So the speed of ore mining increases 2 times for a 64x energy usage increase? Cool.

Techlone:
@Antillar what? Why 64x? I wrote the times for **LV** energy hatch.
For example if you will use IV energy hatch (5th tier) then ODP Mk1 will work 60 ticks; Mk2 - 50 ticks; Mk3 - 40 ticks; Mk4 - 30 ticks.

Antillar:
Ah ok well, I get it. Spare osmium for constant speed and range buff with the same energy usage is very nice.

PR #1214, opened 2017-09-13
Update UO

Technus:
Quality of life changes, also added a piece of code that allows the drain ratio to change with speed multiplier.


ObKo:
So... Oil drilling rig now requires 4x (or 3x?) more energy and have the same rate.... GT oil is pretty useless now. Cycle is 8s (~16k EU at lowest tier) 325L of oil maximum, 50% is <10L, 80% - <100L 92% - <200L. 16k of EU = 80mB of light fuel = 76% of oil sources are useless. IMO, It's almost impossible to find reliable source. Now we have 6 times less (slower) oil than first versions of GT5e, because it already was nerfed by half this year.

PR #1220, opened 2017-09-15
Fix empty oregen

Antifluxfield:
I've just noticed the issue GTNewHorizons/NewHorizons#1909, and this is an alternative fix other than @richardhendricks 's patch.
I mean, why don't we use a dimensional-specified ore vein list? This will cost a bit time to initialize for each dimension, but then it will then speed-up the oregen since it can obviously increase the hit ratio when searching for a random available vein. And the issue should be caused by that the hit failed for too much times when the available vein is occupying a really low proportion in all of the veins.


richardhendricks:
Yes, that would have been my preferred solution.  The empty oregen isn't caused by the hit ratio really, it's more caused by the height limitations.  If an oremix is valid for the dimension, but the chunk doesn't match the height requirements, the chunk stays empty.  My code causes the ore layer to return false if it wasn't able to place any ore at all due to height (or a chunk filled with non-stone, for that matter).

This solution allows us to limit the attempts to known valid oremixes, which will help chunk generation time since we don't have to pass all those variables down to Ore_Layer only to decide not to do it, and the search space for the weighted list is smaller.  And the current 256 limit of attempts can be reduced to something reasonable, since it would directly control how many tries are given to place a valid vein, instead of trying to pick a valid oremix.  

If you check 1909 in 03.log, you can see in Twilight Forest, a valid orevein can take 30+ attempts with valid oremixes to be located (and that there are chunks that give up placing oreveins after reaching the 256 attempt limit). I think a reasonable number would be 4-8 times.  If 4-8 valid oremixes couldn't be placed due to height, too bad so sad you still get an empty chunk, but at least we tried.

Antifluxfield:
Hmmm... I mean what we need to fix is only the empty oregen in that stony world. The height limit can cause a oregen with actually no ore blocks; but I think that's acceptable, because we are used to it and treat it as a feature; but the try-and-fail cycle which exceeds the attempt limit should be an unexpected situation.

richardhendricks:
The stony world empty oregen was an issue with changes that, as far as I know, were only implemented in GTNH GT5U, not mainline Blood-Asp.

Antifluxfield:
Nope. I've already reproduced it:
I disabled most of the veins except for the platinum, and build a superflat world with 250 layers of stones. That is heigh enough for all kinds of vein, so we think that the world should be full of platinum veins:
image
(Ignore that black area. That's where I deleted some stone blocks in-game)
But before we fix it, what it actually generated is like this. And what I attemt to fix is those holes:
image
Like this. No more unexpected empty vein.
Besides, if the vein is cut or erased by the terrian, just let it be. Every experienced GT player have encountered it and know it could really happen.

richardhendricks:
Are those scattered dots small ores?  Because in the GTNH bug, the small ores wouldn't spawn in the chunk either.

Antifluxfield:
Yep.
I may try to find out what happened in GTNH; but that should be another issue.

Techlone:
<h6>

>an issue with changes that, as far as I know, were only implemented in GTNH GT5U

>in the GTNH bug, the small ores wouldn't spawn in the chunk either

Is GTNH even playable? :)
</h6>

Dream-Master:
@Antifluxfield @richardhendricks if you push code to Blood asp branch that fix some thing can you do it for GTNH too or is this allready added ?

richardhendricks:
My changes don't fix anything, just add orevein density control as a variable, and debug messages that can be enabled for the cleanroom.  I'll rebase to the latest GTNH version and provide a pull request for the cleanroom debug.  The orevein changes I want to hold off on until I get a feel for the difference in chunkgen times.

Antifluxfield:
I found that those empty generation without small ores can also be reproduced in this branch, if you move like this:
image
Those 1x1-chunk-sized areas which missing GT ores occurs randomly; but the 4 continuous 3x3-chunk-sized areas in the middle occurs stablely. I'm not really sure how they happened; but I hope the new oregen logic can cover them.

Antifluxfield:
#1227 should be a better fix.

Antifluxfield:
I reopen this repo because of the cascading world gen problem in @Dimach 's pr. Seems that this problem is one of the main reason why we generate the ores only when all 3x3 chunks exists.
I also append some commits for config imporvements simillar to #1227.

PR #1221, opened 2017-09-15
Textures renaming I

Techlone:


Blood-Asp:
I was just about to ask you WTF you are doing. Then i read the 1.10.2...

PR #1226, opened 2017-09-19
Fix AssLine structure check

Antifluxfield:


Dream-Master:
@Antifluxfield did we need that for GTNH too ?

Antifluxfield:
@Dream-Master Yes.

Technus:
Try registering reinforced glass as gt casing?

Technus:
https://github.com/Technus/TecTech/blob/master/src/main/java/com/github/technus/tectech/thing/block/QuantumGlassBlock.java#L92

Just like i do here for my glass

Technus:
https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/api/GregTech_API.java#L294

And here you have explicit warning to not do what u did (does not apply here due to class being read only unles u inject this)

Technus:
Disregard my last post  looks like the only thing that needs to be changed is the addition of the glass to gt api so the state update can propagate properly trough glass (not exacly necessary since glass always touches the proper casing)

Antifluxfield:
@Technus It's ok to register the glass as a casing; but don't you think that the critical point of this issue is just what GregTech_API.java#L294 said?

> You should call @causeMachineUpdate in @Block.breakBlock and in @Block.onBlockAdded of your registered Block.

When placing or breaking the reinforced glass, it doesn't trigger the upgrade. Register the block as a gt casing can only make it possible to transmit an upgrade through it; but an about-to-complete or a just-be-broken AssLine structure can always transmit a triggered upgrade to its controller. So if the glass doesn't call the causeMachineUpdate method, nothing will be fixed.

Technus:
This is why i corrected myself. Only thing that need to be done is to add the casing to gt api.

Technus:
(Add to api on top of your changes to be precise)

Antifluxfield:
That's fine.

Issue #1201, opened 2017-09-03
Crash with Gregtech 5.0.9.31pre4 without forestry installed.


adamcirillo:
Minecraft crashes on start up if forestry isn't installed.

[log.zip](https://github.com/Blood-Asp/GT5-Unofficial/files/1273274/log.zip)

Blood-Asp:
Import of Scoop has to be made optional... will do tomorrw if no one is faster.

PR #1227, opened 2017-09-20
Rewritten oregen, fixed many bugs

Dimach:
Ore veins now generated by chunks, not the entire vein as before.
Its fixed bug when ores not generated sometimes, when its generated in UB stones with wrong textures, etc.
Vein generation changed - veins now have clear boundaries, primary/secondary/sporadic/between system removed, all ores now randomly mixed.
New config for ore generation - removed default parameters from names, dimension configuration now readable. Old config will automatically converted to new file (AdvancedWorldGeneration.cfg). Config file example:
```
gold { //Vein name
            I:Density=3 //Old density parameter
            S:Dimensions < //Dimension block, supports string names or dim ids (working same as oil config)
                Surface
                Moon
                Mars
                Asteroids
             >
            I:MaxHeight=80 //Old max and min height
            I:MinHeight=60
            S:Ores < //New vein contents configuration system - OreID=Weight, count of ores in one vein unlimited
                870=6
                923=1
                86=1
             >
            I:Size=32 //Old size parameter
            I:SizeY=7 //New vein height parameter
            I:Weight=160 //Old weight parameter
        }
```
Much more convenient, right?
Config blocks now can be removed or add with any names, no need to add veins with names like "custom.00".
Asteroids generation moved to separate world gen file, now its can be configured for any dimension.
Ore veins list now dimensional-specified. (fixed bug when ore vein not generated).
This PR conflicts with #1220. In fact, my pr includes the same changes but also implemented new config format and fixed few more bugs.


Antifluxfield:
I think we should disable this generator: https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/loaders/postload/GT_Worldgenloader.java#L21
Or we will have 2 repetitive generators when Galacticraft is installed. That's problematic.

Besides, I prefer that the asteroids should have an individual list other than its dimension list, in order to make its oregen more flexible, enabling some weird idea like adding asteroids with different veins into overworld or something like that... Or just for some other skyland-like dimensions.

mura3277:
If I remember correctly that generator was left there since Galacticraft itself checks for it to enable support?

Antifluxfield:
@Muramasa- So maybe we should switch the generator, instead of adding a redundant one; or just always use that GC-compatible generator only...

Dimach:
@Muramasa- i don't think we still need that class, its used just once, here https://github.com/micdoodle8/Galacticraft/blob/eedc7be8cba5c09aed75ac5f6e3e07a3a2ea4c30/src/main/java/micdoodle8/mods/galacticraft/core/TransformerHooks.java#L193
Galacticraft try to gen field worldGenerators, but this field don't exist anymore. So if delete that class nothing change.

draknyte1:
> veins now have clear boundaries, primary/secondary/sporadic/between system removed, all ores now randomly mixed.  

What do you mean by 'all ore now randomly mixed'?

Dimach:
@draknyte1 before, sporadic and between ores placed in middle 3 blocks of vein. Now they can generate in all vein volume.

draknyte1:
Ahh okay cool. 
Thought it might mean all vein types can spawn mixed randomly. 

I'll give this a review within the next day.

Blood-Asp:
So oreveins are now 1x1 chunks and not up to 5x5? So you have to digg one shaft every chunk instead of every third? Really bad idea...
As the old config is used to generate the new one that is alright, otherwise it would be a no go.
I did not test the GalactiCraft for quite some time now, but did you test it working with your changes? (And that without config changes to GC!)
And why removing the orevein layers? Before you could just digg Top/Middle/Bottom if you only wanted one of the ores in the vein, with that change you need to digg it all absoulutly. Not really happy with that.

Techlone:
>So oreveins are now 1x1 chunks and not up to 5x5?

Wrong. They are still can be 5x5 chunks but GT spawns they chunk by chunk and doesn't set blocks in neighbor chunks.

draknyte1:
I agree with blood that ore layering is nicer. But I chose to set that personal preference aside for what's more realistic. 

If a few people agree though that the mixing is bad, we might have to change that back.

Blood-Asp:
Layering is the realistic way. Different types of stones and ores are found in layers on top of each other in nature.
And i did not yet check the code, but the preview pictures in the first PR looked seperated to me.

draknyte1:
Depends on the ore vein.. finding mixes is definitely a common thing.

Blood-Asp:
Wouldn't that be compount ores that all drop as one type of ore and is seperated in processing?

Dimach:
@Blood-Asp mixing was removed because new config support unlimited count of ores in one vein and adjustable height of vein. If you tell me how ores must be layered with new settings i can implement this.
New generation work almost like old, same sizes, same count of ores. Only differences is clear boundaries and mixed ore. Veins still can be up to 9x9 in size of course.

Antifluxfield:
A compromised solution is to retain the layers but let each layer to generate multiple type of ores.

MauveCloud:
I'm curious: in consideration of "readability" of the config, would it be possible to take a page from GT6's book and use the ore names instead of (or at least as an alternative to) the ids? or would that be too much extra work?

Antifluxfield:
A "material-name to id" map?

leagris:
It already exists within Gregtech and used in `gregtech.lang`.

But a better option would be using the OreDictionary names, or unlocalized names. Like: oreDict → use first matching Gregtech Ore entry with proper container stone.
If unlocalized name → use this specific block.

Example oredict entry:
`ore:oreGold`
Example unlocalized name:
`gregtech:gt.blockores.86.name`

Instead of Numeric ID, would be alternate or better syntax:

```javascript
gold { //Vein name
            I:Density=3 //Old density parameter
            S:Dimensions < //Dimension block, supports string names or dim ids (working same as oil config)
                Surface
                Moon
                Mars
                Asteroids
             >
            I:MaxHeight=80 //Old max and min height
            I:MinHeight=60
            S:Ores < //New vein contents configuration system - "[prefix|ore]:[unlocalizedname|oredictname]"=Weight, count of ores in one vein unlimited
                "ore:oreMagnetite"=6 // Gregtech match on Magnetite ore within appropriate stone type or first oredict if no gregtech ore
                "ore:oreVanadiumMagnetite"=1 // Same as above
                "minecraft:gold_ore"=1 // Specifically use Minecraft Gold Ore regardless of stone type/granite
                "iron_ore"=1 // Match Minecraft Iron Ore as no modid/ore prefix = minecraft
             >
            I:Size=32 //Old size parameter
            I:SizeY=7 //New vein height parameter
            I:Weight=160 //Old weight parameter
        }
```

mura3277:
@Antifluxfield There is one already. Materials.get

NNM42:
can you give dev mod?

NNM42:
2017-09-29_10 57 49_new world_overworld_day

2017-09-29_11 10 48_new world_overworld_day
1. Well.. 5.09.30 - best oregen. IMHO.
2. Why spawn stone only basaltic? (default cfg)
3. More empty chunks.
4. RTG 1.1.1.16, seed -1, centre 0 0, size -1024—1024.
5. SmallOres - very small spawn.
---
so sad.

Antifluxfield:
The game lost its respond when generating a stony superflat world (1 layer of bedrocks and 250 layers of stones)... No idea what's wrong...

NNM42:
need logs

NNM42:
Antifluxfield, yep... inf cycle. "java.lang.StackOverflowError"
Generator opt: 2;7,255x1;1;village
[logs.zip](https://github.com/Blood-Asp/GT5-Unofficial/files/1346033/logs.zip)
WARNING: "fml-client-latest.log" above 1.5Gbyte size.
PS: dependency forestry of GT is normal?  java.lang.NoClassDefFoundError: forestry/api/genetics/IIndividual

Antifluxfield:
@NNM42 You hung up and waited until StackOverflow??!!...
PS:That forestry dependency is already fixed in master branch: #1201

NNM42:
@Antifluxfield i seen this in gregtech.log 
I waited 30-40 min... and open file, next killed proc Java.

Dimach:
@Antifluxfield @NNM42 you must give jvm more memory to generate so large world. I generate that world normally when set -Xmx to 10gb.

Antifluxfield:
@Dimach So you think its normal that you have to generate such world with 10GB? My poor laptop has only 10GB memory, and that's enough for most of those large packs...
The world with same settings generates normally - I mean finish generating within minuites - on the "unstable" branch; but it failed when I recreate the world on this branch. There's definately something wrong...

Dimach:
@Antifluxfield i don't know why my changes make minecraft use so much memory, but when i analyze memory with java visual vm almost all (about 7 gb) was used by chunk instances.

NNM42:
-server -Xms2G -Xmx8G -XX:+UseStringDeduplication.

Dimach:
Okay, after hours of profiling and debugging im finally found memory leak.
Its simple - GT ores have tile entity.
When we set ore block on chunk border world call setTileEntity method which try to notify neighbor blocks, but one of them in other chunk, so world generate another chunk and because in world 230 height all chunks contains veins that process continue while size of generated world not cause OutOfMemory.
I did temporary fix that greatly reduces chance of infinite cycle, but its not fix its cause.
And i know what you think - "Why this problem was not observed at previous generation?", because old world generator had small chance of cause that bug precisely because ore not generated sometimes.

3TUSK:
Seems like another example of [cascading world gen](https://www.reddit.com/r/feedthebeast/comments/5x0twz/investigating_extreme_worldgen_lag/) is here...

Dimach:
New fix of memory leak. Now it fix problem totally - to generate world 230 blocks height minecraft use 2.5Gb.
I used byte code transformer, so code can be hard to understand, but it was tested by several people on several servers (vanila forge, cauldron, thermos, etc) and no problems was found.
Also, to able run coremods from dev client you must use gradle task (runServer or runClient), default IDEA configuration not run it and patch not be applied (i think eclipse have same problem but i not test it). Or just build mod and run it from normal client.

draknyte1:
Your IDE won't load Core mods in dev without program arguments.. That alone gives me my first doubts. I don't like the idea of using Bytecode within Gregtech at all, because unless you're hacking Forge/FML/Vanilla you do NOT need it..

I'll leave @Dragon2488 to review that... for what it's worth

Dimach:
@draknyte1 I'm exactly hacking vanila logic. There is no events that can be canceled, and no flags that can prevent update neighbor blocks. Only way is edit byte code directly.
See yourself - World:2871.
Other way to fix this problem its remove TileEntity from ore blocks, but its almost impossible.

draknyte1:
I am aware of what you're doing (I actually read it after posting) and we will not be merging ASM into GT. Adding special handling in setTileEntity for GT ores isn't something we'll support, it may work but it doesn't fix the actual problem.

Dimach:
New fix by Antifluxfield, without ASM.
Works normally, no performance issues was found.
Please take a look.

MauveCloud:
Here are my feelings on this, FWIW:
1. It's not clear whether this includes the idea about "make sure every 3x3 area has a mix vein", but if so, I hope there's an option to turn that off - I consider finding areas without veins to be part of the challenge of GregTech, though it seems not everyone feels that way.
2. I like the idea of having variable heights for ore veins and more than 4 ore types per mix.  I agree with Blood Asp that layering/strata is better, but I think it could be taken further by making the ore-layer boundaries more "fuzzy".  Maybe that would be too much trouble, though.  More than three "layers" of ores might still be nice though.  OTOH, I could probably endure fully-mixed ore veins if I had to.

Blood-Asp:
How is the current state of this PR? And is the discription up to date?

Dimach:
@Blood-Asp all the bugs that we found fixed. Description up yo date.
The only controversial issue is removed layering of ores. But it can be implemented if needed.

MauveCloud:
There's also the question of whether to use this PR or #1247.  It looks like they both have similar goals in mind, so I'm guessing they'd conflict with each other, meaning it would be better to choose one and close the other.  Or am I misunderstanding something?

draknyte1:
I feel like https://github.com/Blood-Asp/GT5-Unofficial/pull/1247 is that one we will go with.

PR #1230, opened 2017-09-24
Drillers fix

Antifluxfield:
Improve the logic of all drillers;
Rebalance energy cost of Oil Drills.
Now drillers, miners and pumps are posting fake player events when digging or mining.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
+++ b/src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
@@ -205,6 +205,7 @@ public class GT_MetaTileEntity_Miner extends GT_MetaTileEntity_BasicMachine {
 
     protected FakePlayer getFakePlayer(IGregTechTileEntity aBaseTile) {
     	if (mFakePlayer == null) mFakePlayer = GT_Utility.getFakePlayer(aBaseTile);
+    	mFakePlayer.setWorld(aBaseTile.getWorld());
     	mFakePlayer.setPosition(aBaseTile.getXCoord(), aBaseTile.getYCoord(), aBaseTile.getZCoord());
     	return mFakePlayer;
     }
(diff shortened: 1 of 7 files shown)


MauveCloud:
Could you please summarize what the new EU costs and timing will be?  Also, have you actually done anything to make it use the actual max amount from config, rather than effectively half, like that "wrong pumping rate" bug indicates?  Looking at the files changed, I'm having trouble finding where that was fixed.

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Miner.java
```diff
@@ -133,7 +134,9 @@ public void onPostTick(IGregTechTileEntity aBaseMetaTileEntity, long aTick) {
     }
 
     public boolean moveOneDown(IGregTechTileEntity aBaseMetaTileEntity) {
-        if (aBaseMetaTileEntity.getYCoord() + drillY - 1 < 0 || aBaseMetaTileEntity.getBlockOffset(0, drillY - 1, 0) == Blocks.bedrock) {
+        if (aBaseMetaTileEntity.getYCoord() + drillY - 1 < 0 
+        		|| GT_Utility.getBlockHardnessAt(aBaseMetaTileEntity.getWorld(), 0, aBaseMetaTileEntity.getYCoord() + drillY - 1, 0) < 0
```
Dimach: Miners after you changes digs bedrock, so i think you must rewrite this check.

Blood-Asp:
Oil Drill:
T1:	12EU/t		
T2: 48EU/t
T3: 192EU/t
T4: 768EU/t
progress time was doubled from before.

Ore Drill: Same EU/t as oil drill.
progress time was halfed from before.

But as times are dependend on too many called methods or tier based variables, only looking at the PR does not give me the correct numbers. 

Will merge it for all the other changes, the resulting speeds and costs might be balanced again once i have the exact values(maybe ingame)

KayoticEntropy:
@NNM42,  the changes from this pull request are not in the build you have (05.09.31GT) of GT5U.  In fact, no release has been cut since september of last year.  

I'm currently at the stage of my game where I just built an Oil Drilling Rig, and arrived here as I realized: the the stated operation speed is higher than the wiki stated, the amount of oil received per operation is 50% of what was stated during Prospecting, and the cost of the oil field is therefore much lower.  In my case, I was content with a 362L Raw Oil field, but I am only getting 178L per op, and does an op every 8 seconds @ MV (the minimum).  

I'm fine with the speed, but I was hoping to draw the oil much more efficiently with regards to EU.  I like what I believe this change does; would you be able to release a new build of GT5U for 1.7.10?

NNM42:
image

NNM42:
image

PR #1240, opened 2017-10-05
Added TeleporterUsingEvent.

Sirse:
In multiplayer players can teleport to "restricted" areas (like closed dimensions or WorldGuard regions), need to handle it.


draknyte1 (approved):

OvermindDL1:
What the heck is a closed dimension and how does worldedit work from an in-universe perspective (stupid mods that just 'do' things without an in-world explanation should be banned in my opinion...)

However such an event like this would allow someone, say in GT5u, to make an Area Teleport Denial Field (or a Dimensional Teleport Denial Field for an even more utterly massive energy cost).  ;-)

PR #1242, opened 2017-10-09
Oregen fix

Antifluxfield:
I reopen this repo because of the cascading world gen problem in @Dimach 's pr. Seems that this problem is one of the main reason why we generate the ores only when all 3x3 chunks exists.
I also append some commits for config imporvements simillar to #1227.


Antifluxfield:
Nope... There's still something wrong...
Let me see what can I do for the other pr...

Dimach:
@Antifluxfield main problem in my PR its usage of ASM and byte code transformer, wait a little - if current changes will be discarded i rewrite it trying to keep separate chunk generation.

PR #1247, opened 2017-10-15
Oregen Refactor

Antifluxfield:
Based on #1227 .
1. Covered most of the config change, to better parsing the old config and to suit my format. All config modules can be freely extended.
2. Custom ore blocks. Designed to customize the compatibility for mods like GalaxySpace and so on, so you are not suggested to do some strange thing.
`blockores {

    custom01 { #Unlocalized name postfix of this block.
        S:HarvestTool=pickaxe #Harvest tool of this block

        0 { #Subtype 0, the 1st subtype of this block. Maximum 8.
            D:BasicHardness=50.0 #Hardness of this subtype won't be lower than this value.
            I:BlockMeta=0 #Meta ID of the block which this subtype will replace when oregen
            S:Block=minecraft:Obsidian #Name of the block which this subtype will replace when oregen
            S:DustToDrop=Material:Obsidian #The dropped "stone" dust  when harvesting small ores, and the 3rd byproduct in the pulverization recipe. Use "Item:<item name>" for items.
            B:Enabled=false #Set false to disable this subtype.
            S:OrePrefix=ore #OrePrefix of this subtype.
        }
}`

GC & UB blocks are also configurable.
3. Custom stone, small ore, asteroid, and ore vein generation.
Stones:
`        custom_stone {
            I:BlockMeta=0 #Meta id of the block of this stone
            S:Block=NULL #Block name of the block of this stone

            custom_scale {
                B:AllowedToGenerateInVoid=false
                I:Amount=1 #Amount of stones the generator will try to generate. Suggest to be 1.
                S:DimensionWhiteList < #Dimension white list. Empty list meas won't generate in any dimension.
                    Overworld #Dimension name
                    ?End #Match the dimensions whose name contains "End"
                    @Hell #Match the dimensions whose class name contains "Hell"
                    -28 #Dimension ID
                 >
                I:MaxY=0
                I:MinHeight=0
                I:Probability=0
                S:RestrictedBiomes < #Biome white list. Unlike the dimension list, empty list means no restriction.
                 >
                I:Size=0
            }

        }`
Small ores:
`small {

            custom {
                I:Amount=0
                S:DimensionWhiteList <
                 >
                I:MaxHeight=0
                I:MinHeight=0
                S:Ore=NULL #Material name of the small ore. Support the numerical id.
                S:RestrictedBiomes <
                 >
            }
}`

Asteroids:
`    asteroid {

        custom_asteroid { #Name of the asteroid
            I:BlockMeta=0 #Meta id for the block of the asteroid
            S:Block=NULL #Block name for the block of the asteroid
            S:DimensionWhiteList <
             >
            I:MaxHeight=0
            I:MaxSize=0
            I:MinHeight=0
            I:MinSize=0
            I:Probability=0
            S:RestrictedBiomes <
             >
        }
}`

Ore vein:
`        mix {

            custom {
                S:Asteroids < 
                 >#The name of the asteroids above. The asteroids will use their own dimension white list, so this list only tells the asteroid which vein they can use.
                I:Density=0 #Now density argument will also affect asteroids.
                S:DimensionWhiteList <
                 >
                I:MaxHeight=0
                I:MinHeight=0
                S:OrePrimaryLayer <#Ores in each layer
                    Material:Copper::60 #GT copper ore, with weight 60 in this layer.
                    Material:Iron::30 #GT Iron ore, with weight 30 in this layer.
                 >
                S:OreSecondaryLayer <
                    Material:67 #GT neodymium ore, with default weight 100 in this layer
                 >
                S:OreSporadiclyInbetween <
                    Block:minecraft:restone_block::45 #Vanilla redstone block, weight 45
                 >
                S:OreSporaticlyAround <
                    Block:dirt:1::30 #Vanilla dirt block with meta 1, weight 30
                 >
                I:RandomWeight=0
                S:RestrictedBiomes <
                 >
                I:Size=0
                I:Thickness=7 #Thickness of the vein. Minimum 5.
            }
}`


Techlone:
Why does your config have default values in names? Stop it.

Antifluxfield:
@Techlone It can be considered to remove those default values; but they are just added by the old getters in GT_Config. Most of our values are readed through these getters, thus I do so as well.
Or you mean we should renovate all our config files to remove all those defaults?

Techlone:
>Or you mean we should renovate all our config files to remove all those defaults?

Not in this PR. Just do not use these getters.

Antifluxfield:
@Techlone OK... will change it.

Blood-Asp:
I certainly need some to test this before merging something soo huge. 

Also about the removal of the special Galacticraft ore handler: Are you certain it now works without? And also without enabling other mods generation in GC config as this is certainly not wanted and breaks a lot?

richardhendricks:
So ores will be spread throughout the whole vein? I don't care for that. Having the ores located in a certain part of the vein helps with mining a targeted ore.

src/main/java/gregtech/common/blocks/GT_TileEntity_Ores.java
```diff
@@ -41,7 +41,7 @@ public static boolean setOreBlock(World aWorld, int aX, int aY, int aZ, int aMet
         return setOreBlock(aWorld, aX, aY, aZ, aMetaData, isSmallOre, false);
     }
 
-    public static boolean setOreBlock(World aWorld, int aX, int aY, int aZ, int aMetaData, boolean isSmallOre, boolean air) { //TODO
+    public static boolean setOreBlock(World aWorld, int aX, int aY, int aZ, int aMetaData, boolean isSmallOre, boolean air) {
         if (!air) {
```
richardhendricks: Need to redo the air boolean logic.  It does not actually enable what you would think, ie placing ores in air. It also causes problems with placing ores above Y=128 in the Deep dark since getActualHeight does not correlate with the real max height of the dimension.  getActualHeight is based on hasNoSky(which is a terribly incorrect name, it should instead be called isNether or smallHeight).  My suggestion is to remove the air boolean completely. Validating ores place at legal heights should be done when reading the config file, or not at all (my preference).

richardhendricks (changes requested):

Blood-Asp:
too many issues, and no more fixes. so dead...

PR #1249, opened 2017-10-16
Fix unexpected recipe loading

Antifluxfield:
Fix unexpected recipe without forestry:
image


draknyte1 (approved):

Blood-Asp:
Best would maybe that re recipes does not load at all with missing inputs, but there is becomes complicated as it transmits all free input slots as null by default..

PR #1251, opened 2017-10-17
Localization improve

Antifluxfield:
Make tooltips of tools able to be localized
Improve the mechanism of i18n of ores and most meta-generated items.


draknyte1 (approved):

Blood-Asp:
Reminds me that there is still a problem with translation that is does not really seems to work with that i did at my last change to it.

PR #1252, opened 2017-10-17
Configed metal blocks

redmage17:
Added config for removal of shapeless recipes for metal blocks. Tested it works.


redmage17:
It allows you to config whether or not your need to use the compressor to create metal blocks from ingots.

draknyte1 (changes requested):
So, a few things.
Bad use of a Scala import.
mEasierCasings defaults to true, should be false.
Youve changed small boiler automation, thats not going to be accepted. Make it a config option if you want it flexible.

I prefer you make each change as a seperate PR instead of bundling them together.

redmage17:
I didn't realize the pull request would update as I worked with the fork have been testing a few things, let me rewrite this

redmage17:
Also where is the bad Scala import? I can't find any changes that I made that should use Scala to my knowledge.

redmage17:
Issues look to be fixed

tomasbrod:
I suggest you to either create a new PR or remove the unwanted commits from this BRANCH (usign git reset --hard, or rebase -i) and push --force.

PR #1255, opened 2017-10-20
ItemPipe tick time customization

Antifluxfield:
Fix bugs & tooptips


draknyte1 (changes requested):
Dont change the constructor.. add a new one..
I dont see the point in this though?

Antifluxfield:
Generally, nothing was changed, since we've never used the constructor with the time of ticking cycle as an argument. But if anyone use it, the tile won't be created as expected in the world.

draknyte1:
I dont see why we ever would..
No harm done having extra flexibility though.

Antifluxfield:
Also no idea why we have that constructor... But since it's already there, I think it would be better to fix it.

PR #1267, opened 2017-11-01
Configed Storage Blocks

redmage17:
Added config for removal of shapeless recipes for storage blocks


draknyte1 (approved):
Much cleaner.

redmage17:
Getting better at this

PR #1271, opened 2017-11-04
Plantball fix

jaskarth:
Re add the recipe for making plantballs from all saplings (should work with all saplings registed with `treeSapling`)

### DIFF
--- a/src/main/java/gregtech/loaders/oreprocessing/ProcessingSaplings.java
+++ b/src/main/java/gregtech/loaders/oreprocessing/ProcessingSaplings.java
@@ -1,6 +1,7 @@
 package gregtech.loaders.oreprocessing;
 
 import gregtech.api.enums.GT_Values;
+import gregtech.api.enums.ItemList;
 import gregtech.api.enums.Materials;
 import gregtech.api.enums.OrePrefixes;
 import gregtech.api.util.GT_ModHandler;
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/loaders/oreprocessing/ProcessingSaplings.java
```diff
@@ -15,6 +16,7 @@ public ProcessingSaplings() {
 
     public void registerOre(OrePrefixes aPrefix, Materials aMaterial, String aOreDictName, String aModName, ItemStack aStack) {
         GT_ModHandler.addPulverisationRecipe(GT_Utility.copyAmount(1L, new Object[]{aStack}), GT_OreDictUnificator.get(OrePrefixes.dustSmall, Materials.Wood, 2L), null, 0, false);
+        GT_ModHandler.addCompressionRecipe(GT_Utility.copyAmount(8L, new Object[]{aStack}), ItemList.IC2_PlantballCompressed.get(1L, new Object[0]));
```
MauveCloud: ItemList.IC2_PlantballCompressed? Isn't that "Compressed Plants" which isn't used anymore in IC2Exp? (see https://wiki.industrial-craft.net/index.php?title=Compressed_Plants )

jaskarth:
I saw some other recipes using that so I decided to use that as well. I'm testing it without the compressed part and I'll see what happens.

jaskarth:
Yes it still works; thanks for letting me know.

PR #1272, opened 2017-11-04
Magic fix

jaskarth:
Not really a fix but rebalancing to make it more effective on small nodes.
The reason why it was not generating in the old version was because the energy from vis divisor was set way too high, meaning a node with 60 or so of each aspect was necessary to generate 32 eu per tick. Also, the energy will scale better now, instead of a huge amount of power given when you get more than 4 of each aspect on a node, a small boost will be given to each aspect. This should make it more viable for practical use.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
+++ b/src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
@@ -1,5 +1,6 @@
 package gregtech.common.tileentities.generators;
 
+import cpw.mods.fml.common.FMLLog;
 import cpw.mods.fml.common.Loader;
 import gregtech.api.GregTech_API;
 import gregtech.api.enums.ConfigCategories;
@@ -38,11 +39,13 @@ public class GT_MetaTileEntity_MagicalEnergyAbsorber extends GT_MetaTileEntity_B
     public static boolean sAllowMultipleEggs = true;
     public static GT_MetaTileEntity_MagicalEnergyAbsorber mActiveSiphon = null;
     public static int sEnergyPerEnderCrystal = 32;
-    public static int sEnergyFromVis = 12800;
+    public static int sEnergyFromVis = 512; //really? this high? no wonder no eu was generating... (it was 12800 at the time)
     public static int sDragonEggEnergyPerTick = 128;
     public static boolean isThaumcraftLoaded;
+    public static boolean tLameMode = false; // if you use this... lame.
     public int mEfficiency;
     public EntityEnderCrystal mTargetedCrystal;
+    //public static int[] VoltageOut = {32, 128,512,2048}; // Because what the fuck even is this code.
 
     public GT_MetaTileEntity_MagicalEnergyAbsorber(int aID, String aName, String aNameRegional, int aTier) {
         super(aID, aName, aNameRegional, aTier, "Feasts on magic close to it", new ITexture[0]);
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
```diff
@@ -38,11 +39,13 @@
     public static boolean sAllowMultipleEggs = true;
     public static GT_MetaTileEntity_MagicalEnergyAbsorber mActiveSiphon = null;
     public static int sEnergyPerEnderCrystal = 32;
-    public static int sEnergyFromVis = 12800;
+    public static int sEnergyFromVis = 512; //really? this high? no wonder no eu was generating... (it was 12800 at the time)
```
Exidex: I don't think leaving comments like this is a good thing. If you want something to be changed, just change it. Reasoning behind this put into PR description.

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
```diff
@@ -38,11 +39,13 @@
...
     public static int sDragonEggEnergyPerTick = 128;
     public static boolean isThaumcraftLoaded;
+    public static boolean tLameMode = false; // if you use this... lame.
```
Exidex: What even is this variable? How do you set it?

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
```diff
@@ -38,11 +39,13 @@
...
     public int mEfficiency;
     public EntityEnderCrystal mTargetedCrystal;
+    //public static int[] VoltageOut = {32, 128,512,2048}; // Because what the fuck even is this code.
```
Exidex: What is this doing here?

Exidex (changes requested):

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
jaskarth: Ugh I'm really sorry I left those things in when I was testing and forgot to remove them. I will fix those as soon as possible.

jaskarth:
I'm really sorry again, I will clean up my code better before pull requesting next time.

Exidex (approved):

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
```diff
@@ -109,34 +110,31 @@ public void onPostTick(IGregTechTileEntity aBaseMetaTileEntity, long aTick) {
             // Energyzed node
             if (isThaumcraftLoaded) {
                 try {
+                	int multFactor = 2;
                     World tmpWorld = this.getBaseMetaTileEntity().getWorld();
                     int tmpX = this.getBaseMetaTileEntity().getXCoord();
-                    int tmpY = this.getBaseMetaTileEntity().getYCoord();
+                    int tmpY = this.getBaseMetaTileEntity().getYCoord() + 2; //Add 2 because +1 is the node stabiliser and +2 is the node itself
```
Blood-Asp: These coordinates are of the energy absorber. We have no idea where the node or any thaumcraft parts are. We only tell the TC API that here is something that wants to consume Vis and it pushes it there if in range.

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
```diff
@@ -109,34 +110,31 @@ public void onPostTick(IGregTechTileEntity aBaseMetaTileEntity, long aTick) {
...
-                    if (air > 4)
-                        mult += 15;
-                    if (destruction > 4)
-                        mult += 15;
-                    if (order > 4)
-                        mult += 15;
-                    if (water > 4)
-                        mult += 15;
-                    visEU = (visEU * mult) / 100;
-                    getBaseMetaTileEntity().increaseStoredEnergyUnits(Math.min(maxEUOutput(), visEU * getEfficiency() / this.sEnergyFromVis), false);
```
Blood-Asp, line 138: the value returned from "getEfficency()" is 10000 for 100% i somehow get the feeling that the changed EnergyFromVis value does not account for that.

jaskarth:
I wasn't aware the TC API worked like that, the latest commit should fix that.

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
jaskarth: doesn't the getEfficiency() function return the efficiency of that block? Such as 90% for LV, 80% for MV, 70% for HV, and so on? I was trying my best to keep the old code's functionality so I just left it there like that

Blood-Asp: looking at it again is seems you are right. was working too much with the multiblocks lately to remember that single blocks work differently...

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
```diff
@@ -1,5 +1,6 @@
 package gregtech.common.tileentities.generators;
 
+import cpw.mods.fml.common.FMLLog;
```
draknyte1: ?

draknyte1 (changes requested):
Useless Import.

src/main/java/gregtech/common/tileentities/generators/GT_MetaTileEntity_MagicalEnergyAbsorber.java
jaskarth: No idea how that got there

draknyte1 (approved):

PR #1276, opened 2017-11-05
Gasoline

jaskarth:
I feel that having more ways of making electricity from oil is beneficial, so I decided to add gasoline. Gasoline has to 3 main forms: raw gasoline, regular gasoline, and premium gasoline. Raw gasoline is made by combining light fuel, refinery gas, acetone, and ethanol. Raw gasoline cannot be burned as it would cause to much engine knock; In order to use it you must add lead to it to reduce the engine knock. This creates "regular" gasoline, basically what you buy in a gas station (In the United States at least), and it makes 384,000 EU. In order to make "premium" gasoline (the stuff that no one can afford) you mix in octane, toluene, and nitrous oxide to "regular" gasoline. This makes 640,000 EU. Octane is produced by distilling hydro-cracked light fuel. All suggestions on how to make this better welcome.


MauveCloud:
> Raw gasoline cannot be burned as it would cause to much engine knock; In order to use it you must add lead to it to reduce the engine knock. This creates "regular" gasoline, basically what you buy in a gas station (In the United States at least)

Um... I thought we switched over to "unleaded" gasoline for our cars a few decades ago. 😁

jaskarth:
I changed lead with toluene, no more poison :) however it's going to make gasoline a bit more expensive

JohannesGaessler:
I think the starting material should be Naphtha instead of Light Fuel.
I think that this would make more sense in terms of gameplay since players would then have a choice between producing more plastic and processing Naphtha into Gasoline.
I don't think it would be necessary to have Raw Gasoline as an intermediary then.

Also, I think you might be confusing nitric oxide (NO) with nitrous oxide (N2O), see:
https://en.wikipedia.org/wiki/Nitric_oxide
https://en.wikipedia.org/wiki/Nitrous_oxide

I think that for Premium Gasoline it might make more sense to use ethyl tert-butyl ether instead of octane as the anti-knocking additive.
It could be produced from Ethanol and Butene.
https://en.wikipedia.org/wiki/Ethyl_tert-butyl_ether
I would also suggest renaming Premium Gasoline to Octane-Boosted Gasoline so it's more clear what's being modeled.

src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java
```diff
@@ -3368,6 +3369,10 @@ private void addRecipesMay2017OilRefining() {
 				new FluidStack[]{Materials.LightFuel.getFluid(100), Materials.Naphtha.getFluid(125), Materials.Toluene.getFluid(80), Materials.Benzene.getFluid(400), Materials.Butene.getGas(80), 
 						Materials.Butadiene.getGas(50), Materials.Propane.getGas(10), Materials.Propene.getGas(100), Materials.Ethane.getGas(15), Materials.Ethylene.getGas(150), Materials.Methane.getGas(150)}, 
 				Materials.Carbon.getDustTiny(3), 120, 120);
+		//Recipes for gasoline
+				GT_Values.RA.addMixerRecipe(Materials.LightFuel.getCells(16), Materials.Gas.getCells(2), Materials.Ethanol.getCells(1), Materials.Acetone.getCells(1), GT_Values.NF, GT_Values.NF, Materials.GasolineRaw.getCells(20), 200, 120);
+				GT_Values.RA.addChemicalRecipe(Materials.GasolineRaw.getCells(10), Materials.Toluene.getCells(1), GT_Values.NF, GT_Values.NF, Materials.GasolineRegular.getCells(11), 10, 120);
+				GT_Values.RA.addMixerRecipe(Materials.GasolineRegular.getCells(20), Materials.Octane.getCells(2), Materials.NitricOxide.getCells(6), Materials.Toluene.getCells(1), GT_Values.NF, Materials.GasolinePremium.getFluid(25), ItemList.Cell_Empty.get(29), 200, 120);
```
JohannesGaessler: I think the output is supposed to be 25000L of Premium Gasoline instead of 25L.

jaskarth: Fixed the typo

jaskarth:
I forgot that naptha existed, lol. I'll go change that. I think that Raw Gasoline should stay, because that would make gasoline production more complex and require more industry, as gasoline produces a lot of power (384K/1000L regular 640K/1000L octane-boosted) and being able to produce that amount of power in the earlygame would unbalance it heavily. That nitrous oxide thing was a typo. I want to make it so both ethyl tert-butyl ether and octane will be used. I can also boost the output to 768 thousand EU, as making ethyl tert-butyl ether will require an hv chemical reactor, making octane boosted gasoline an HV fuel.

PR #1279, opened 2017-11-10
Resistor recipe change

JohannesGaessler:
Currently Resistor recipes only accept Coal Dust.
I think this is unneccarily restrictive.
This PR would allow to alternatively use Charcoal Dust or Carbon Dust.

I also removed the crafting recipe that uses 1x Wires instead of Fine Wires since the recipe is objectively worse and only serves as a noob trap.


tomasbrod:
Why not graphite dust?

MauveCloud:
> Why not graphite dust?

Maybe because you can already electrolyze graphite dust for 4 carbon dust.  (you can also electrolyze coal dust for 2 carbon dust, but it's somewhat less clear whether that's worth doing)

Barhandar:
Don't make unrelated changes (wire-using recipe removal) in the same commit.

JohannesGaessler:
@Barhandar The change is not unrelated.
Currently there are 2 crafting recipes for Resistors: [Coal Dust] x [1x Wire, Fine Wire].
Adding Charcoal Dust and Carbon Dust without further changes would have increased the number of recipes to 6 and only further increased the likelihood of confusion.

Barhandar:
The change is unrelated. PR deals with adding carbon and charcoal dusts to the recipe, not with your ideas about how more crafting actions is an acceptable and unquestionably superior way over using slightly more resources, and that the latter should not be an option. And said change is not announced in the commit, bordering on _trying to sneak it in_.
>Adding Charcoal Dust and Carbon Dust without further changes would have increased the number of recipes to 6 and only further increased the likelihood of confusion.

Register wire-fine-wire-fine-wire-fine, every page will then have single dust type, preventing confusion from all but the most dim players - ones which you're implying are the majority, and which ones are actively filtered by other circuit changes anyway.

JohannesGaessler:
>trying to sneak it in.

If I was trying to sneak things in I wouldn't be making a PR in the first place.
I also wouldn't explicitly state that I did it in the OP.

>the most dim players - ones which you're implying are the majority

Please stop putting words into my mouth.

Anyways, the possible confusion comes from there being a recipe with 1x Wire in the first place.
The implication of having two recipes like this is that one recipe needs you to craft more machines and in turn uses up less resources, like how you can make Plates in a machine at half the cost compared to crafting them with a Hammer.
For example, when I first played with the new Circuits I didn't know that you could make Fine Wires without a Wiremill.
This belief was only further reinforced by having two recipes that are identical except for the Copper cost.
If there had only been the recipe with Fine Wires it would have been obvious that you'd have to be able to craft them even without machines.

Both recipes become available at exactly the same time.
The fundamental metric for how expensive something is in GT is how long it takes the player to achieve it.
The amount of time it takes to Hammer a bunch of Plates into Foils is absolutely miniscule compared to the amount of time it takes to mine Copper.
Therefore the recipes with Fine Wires are objectively better.

Issue #1282, opened 2017-11-11
Fix for uselessness of Oil Cracking Unit


Antillar:
Currently Oil Cracking Unit is useless for the folowing reasons:
1. Cracking can be done in Chemical Reactor.
2. Cracking can be done in LCR.
3. This machine occupies almost twice as much space as LCR.
4. This machine is more expensive than a Chemical Reactor at LV or MV or HV (the machine is HV-tier).
5. There are no losses in production values when cracking in any other machine.

The last reason is the one I suggest to change. 
As the OCU is the machine specially designed for cracking hydrocarbons, it should be more efficient than other machines at this job.

I suggest nerfing the amount of resulting fluid to 70% or 80% of the non-cracked fluid when cracking NOT in OCU. 

Bear has suggested another solution, which does not fully cover the problem in my opinion, but also worth considering. This is the link to his issue: https://github.com/Blood-Asp/GT5-Unofficial/issues/1268

JohannesGaessler:
>I suggest nerfing the amount of resulting fluid to 70% or 80% of the non-cracked fluid when cracking NOT in OCU.

Cracking in a (Large) Chemical Reactor currently runs at 80% efficiency.

Issue #1268, opened 2017-11-02
complaint, idea for cracking


bear989:
the cracking unit should be the best way for cracking but as it is there is a 200L gain for using it at a 30K eu cost and a lot of stainless to make in the first place. all levels of cracking can now be done in the chemical reactor at LV but the cracking unit takes HV. the idea witch is from netmc is that if your gonna do all levels of cracking in LV then you should have to do it in steps, so you would need to run 3 reactors to get the severely cracked that gives the best ethylene witch is probably what we are wanting from the cracking and that would get the power usage up as well. my suggestion is to only allow the severely and moderately cracked in the cracking unit, save the best outputs for the best option to use.

jaskarth:
yes and we can up the cracking to needing mv, as well

mura3277:
Assuming this was fixed by cracking followup 2

PR #1283, opened 2017-11-11
Cracking followup 2

JohannesGaessler:
Currently there are 11 Materials that can be cracked, resulting in 66 recipes with 6 recipes per material.
However, for several materials this level of complexity is not necessary and the amount of infrastructure needed for cracking is too expensive compared to the benefits.

This PR removes Cracking for the Following Materials:

* ~~Refinery Gas~~
* Ethylene
* Propene
* Butene
* Butadiene

Instead this PR adds the following conversions:

* Ethane <-> Ethylene + Hydrogen
* Ethylene <-> Propene
* Propane <-> Propene + Hydrogen
* Ethylene <-> Butene
* Butane <-> Butene + Hydrogen
* Butene <-> Butadiene + Hydrogen

These recipes would also allow the (inefficient) production of plastics from Natural Gas.

This PR also reduces the processing time of the Oil Cracking Unit by a factor of 2.

I also fixed a bug where I forgot to increase the processing time of cracking in a (Large) Chemical Reactor in an earlier PR.

I also increased the yield of cracking Heavy Fuel, Light Fuel, and Naphtha.
It seems I made a mistake when I calculated the appropriate yields for those materials, which resulted in the yields being too low.
For instance, cracking Heavy Fuel into Light Fuel did not actually yield net energy compared to just burning the Heavy Fuel in a Large Boiler.


Antillar:
I dislike that refinery gas can no longer be cracked. Steam-cracking of refinery gas was a useful way of getting the useful hydrocarbons. I do not really get the point of moderate cracking, but light cracking gave 2566 methane, 213 ethylene and 113 propene, all of them being useful, and the only useless material was ethane, which was only obtained in the amount of 19 mb per operation. Severe cracking gave maximal methane output - 914 methane and only 48 not methane. Additionally, this was a good source of helium used for crystalprocessors and Nq alloy without the need to get to the End and mine stone.

I aspecially like the ability to create butadiene (SBR) from useless butene 1:1

I may be mistaken, but I have not seen butane obtained from steam-cracking completely, also not seen any direct usage of it.

Related issues: https://github.com/Blood-Asp/GT5-Unofficial/issues/1268 https://github.com/Blood-Asp/GT5-Unofficial/issues/1282

JohannesGaessler:
@Antillar Right, I didn't consider that useage of that recipe.
I'll proably reenable it.

Antillar:
I also suggest removing the naphta, light fuel and heavy fuel outputs from distilling cracked naphta, light fuel and heavy fuel. This is unnecessary recursion IMO. I don't see the need of obtaining one from another, as none of them can be used for something special (except naphta being epoxy component in remaining deprecated recipe). The severity of cracking wisely chosen on the following stage can totally neglect the unusual requirement of special hydrocarbons obtained from each of these oil fractions.

Another thing I would like to see is the ability to recycle **acetone** to other hydrocarbons. I have found this: "On distillation with sulphuric acid (H2SO4), it is converted into mesitylene C9H12 (symmetrical trimethyl benzene)." [(source)](http://www.worldofmolecules.com/solvents/acetone.htm)
And that hydrocarbon can further be broken down to useful ones, likely, benzene and methane?
The reason behind this is that for every mb of epoxy produced there is 1 mb of acetone remaining unused. This allows to recycle it further.

JohannesGaessler:
I don't think the feedback loops between Naphtha, Light Fuel, and Heavy Fuels should be removed.
They incentivize building more infrastructure to get better efficiency in terms of Oil.

>The severity of cracking wisely chosen on the following stage can totally neglect the unusual requirement of special hydrocarbons obtained from each of these oil fractions.

I don't understand what you're trying to tell me here.

>On distillation with sulphuric acid (H2SO4), it is converted into mesitylene C9H12 (symmetrical trimethyl benzene)

I think that solution would be overly complicated.
Do you think it would be enough to allow the burning of Acetone in a Gas Turbine?

Antillar:
I think that sorting output from distilled cracked naphta, LF and HF and putting these 3 products back into OCUs does not make the overall infrastructure bigger, only adding extra piping required to connect both
3,3-graphs. 
In my particular case I store all fluids in ME network, so I really do not care where the fluid came from, I just export it to a certain machine.

The reason I am against excessive acetone in epoxy chain is that acetone can only be converted into useless glue or cetane-boosted diesel, and first one can be obtained from LV centrifuge processing Sticky Resin, while the second one is unlikely used by any player, because it simply burns precious oil, which is so expensive to fully process into resins. I admit that once in a while someone appears to prove me wrong, but these are mostly temporary setups, originated from the lack of energy in one place.

What I do with acetone is to break it into hydrogen and carbon, but this way for each mb of epoxy processed from hydrocarbons I waste a whole 1 mb of hydrocarbons as acetone. So I was trying to suggest a way to recycle acetone into hydrocarbons, so the epoxy chain is wasteless.

JohannesGaessler:
Regarding feedback:
The minimal setup for Cracking with a Distillation Tower is to distill Steam-Cracked Naphtha.
If you then get more Light Fuel/Heavy Fuel from that process it gives you a synergy with processing those other Oil products.
You're cracking everything anyways from what I remember from your screenshots.

Regarding Acetone:
What if I were to add more recipes that let you specifically produce Phenol (from Benzene)?

Antillar:
I think, when phenol can be produced separately, the player can produce 2x phenol + 2x acetone by current recipe and 1x separate phenol, and then turn it into wasteless epoxy. Sounds great.

PR #1290, opened 2017-11-19
[1.10.2]Compressed Block registry rework

Antifluxfield:
Why you are registering those compressed blocks with the algorithm highly depending on the order of the materials?

### DIFF
--- a/src/main/java/gregtech/common/blocks/MetaBlocks.java
+++ b/src/main/java/gregtech/common/blocks/MetaBlocks.java
@@ -3,6 +3,7 @@ package gregtech.common.blocks;
 import gregtech.api.capability.internal.IGregTechTileEntity;
 import gregtech.api.metatileentity.IMetaTileEntity;
 import gregtech.api.metatileentity.PaintableMetaTileEntity;
+import gregtech.api.unification.material.Materials;
 import gregtech.api.unification.material.type.DustMaterial;
 import gregtech.api.unification.material.type.Material;
 import gregtech.api.unification.ore.OrePrefix;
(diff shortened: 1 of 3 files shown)


Exidex (changes requested):
Fix your whitespaces and formatting, also there is RegistryNamespaced#getIDForObject method that does exactly the same as the one you added

Antifluxfield:
Done.

Exidex (commented):
So, going to the point. I am not sure if this change is necessary. Existing code and proposed, both have a requirement not to change material ids order. The only difference I see is if we'll want to make material 'not ignored' (I don't think will want to do that) or change a material type, this will not break backward compatibility, but this is at the cost of increasing amount of blocks created. Also this code will change in 1.13 because of vanilla changes.

Antifluxfield:
You should know that the order of the materials are always more volatile than the fixed ids. Imagine someone register his own material through the configs or by an addon:
Suppose our current materials is occupying the id segment O, an addon registered its materials to id segment A, and we append some materials once in an update in id segment B. Assume that A is following B, If a player takes the addon before taking the update, the order of the materials will change form "O->A" to "O->B->A", and breaking the backward compatibility. Same thing will happen if A lays before B and the player takes the addon after the update.
The only way to solve this issue is to patch a config to adjust the ids for each case when you want to append some material, and let the players/modpackers to set them to a proper value - I don't think that's a good idea since we can fix it know.

src/main/java/gregtech/common/blocks/MetaBlocks.java
```diff
@@ -106,36 +107,40 @@ public static void init() {
...
         int generationIndex = 0;
         for(Material material : Material.MATERIAL_REGISTRY.getObjectsWithIds()) {
             if(material instanceof DustMaterial) {
-                materialBuffer.add((DustMaterial) material);
-                if(materialBuffer.size() == 16) {
-                    createCompressedBlock(materialBuffer, generationIndex);
-                    materialBuffer.clear();
-                    generationIndex++;
-                }
+            	int id = Material.MATERIAL_REGISTRY.getIDForObject(material), index = id / 16;
```
Exidex: I spend some time looking where index was declared. Separate this line into two and we are good to merge

Exidex (changes requested):

Issue #1278, opened 2017-11-07
Suggestion: Flexibility for oil crackers


Antillar:
Currently Oil Cracking Unit (OCU further) requires hydrocarbons to be inserted and extracted from sides (left and right), this is ok. H2O^ and H2^ can be inserted in the middle ring. This also makes sense. But what else does a player need to fit in the middle ring?
* Controller
* Maintenance Hatch
* Energy Hatch
* and adding previously mentioned Input Hatch

Overall there are 4 neccessary interactive blocks in the middle ring, which by itself only contains 8 blocks.
The side rings, however, have only 1 Hatch on them out of 9 blocks they consist of.

I suggest allowing more flexibility for OCUs in the following details:
* Allow Maintenance hatch to be placed on side ring of casings.
* Allow Energy Hatch to be placed on side ring of casings.

What this allows:
* Players can now have only 2 necessary blocks on each ring: Input Hatch/Output Hatch + Maintetance Hatch/Energy Hatch on one side ring and the rest on the other, Controller and Input Hatch for Hydrogen/Steam on middle ring.
* All I/O blocks along with Maintenance Hatch and Controller can be placed on 1 Y level - the middle level. And they can be placed so they do not touch each other. Also they can be placed to face only front and back sides of the machine (the ones 5 blocks long).
* Due to these blocks placable all in one heigth level, the OCUs can be independently stacked vertically, as top and bottom layers will contain none of interactive blocks, and therefore the machines are independent as they only share coil blocks and casings.

bear989:
I disagree as there is plenty of room in that center block for all you need there why change it? The center is where the work is being done so that is where the power needs to be and maintenance would be needed.

Blood-Asp:
One thing to add for flexibility would likely be processing direction. so that input/output can be placed on either end, but only one a time. I just left that out when adding the multiblock because the code already became quite complex.

PR #1297, opened 2017-11-22
More flexible Oil Cracker structure

JohannesGaessler:
Currently the structure of the Oil Cracker is unneccessarily restrictive.
You currently have to put the Input Hatch on the left hand side, an the Output Hatch on the right hand side.
This PR would make it so that the Oil Cracker also accepts Input Hatches on the right hand side and Output Hatches on the left hand side.
Input/Output Hatches must still be on opposite sides of one another.

Also, Energy/Maintenance Hatches no longer need to be on the middle ring but can also be put on the sides.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_OilCracker.java
+++ b/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_OilCracker.java
@@ -38,12 +38,13 @@ public class GT_MetaTileEntity_OilCracker extends GT_MetaTileEntity_MultiBlockBa
                 "Thermally cracks heavy hydrocarbons into lighter fractions",
                 "Size(WxHxD): 5x3x3 (Hollow), Controller (Front center)",
                 "Ring of 8 Cupronickel Coils (Each side of Controller)",
-                "1x Hydrocarbon Input Bus/Hatch (Any left side casing)",
+                "1x Hydrocarbon Input Bus/Hatch (Any left/right side casing)",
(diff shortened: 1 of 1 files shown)


Antillar:
If only I could like it more than once.

src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_OilCracker.java
```diff
@@ -38,12 +38,13 @@ public GT_MetaTileEntity_OilCracker(String aName) {
...
                 "Size(WxHxD): 5x3x3 (Hollow), Controller (Front center)",
                 "Ring of 8 Cupronickel Coils (Each side of Controller)",
-                "1x Hydrocarbon Input Bus/Hatch (Any left side casing)",
+                "1x Hydrocarbon Input Bus/Hatch (Any left/right side casing)",
                 "1x Steam/Hydrogen Input Hatch (Any middle ring casing)",
-                "1x Cracked Hydrocarbon Output Hatch (Any right side casing)",
-                "1x Maintenance Hatch (Any middle ring casing)",
-                "1x Energy Hatch (Any middle ring casing)",
-                "Clean Stainless Steel Machine Casings for the rest (18 at least!)"};
+                "1x Cracked Hydrocarbon Output Hatch (Any left/right side casing)",
```
Antillar, line 43: right/left is more logical since they are on the opposite sides and the first one was left/right

Antillar:
https://github.com/Blood-Asp/GT5-Unofficial/issues/1278

Issue #1277, opened 2017-11-06
Suggestion: allow vertical stacking for LCRs


Antillar:
Currently LCR does only allow horizontal stacking, left-to-right and wise versa. This is because of 1 Cupronickel Coil Block at the bottom and none on top. I suggest adding another Cupronickel Coil Block on top. 

With this small addition the LCRs will finally allow vertical stacking, as top Cupronickel Coil Block for bottom LCR can serve as a bottom Cupronickel Coil Block for top LCR. 

Many players including me prefer stacking machines vertically, and this small addition allows us to also do it with LCRs.

Antillar:
https://github.com/Blood-Asp/GT5-Unofficial/pull/1298

PR #1298, opened 2017-11-22
More Flexible Large Chemical Reactor structure

JohannesGaessler:
Currently it is not possible to vertically stack Large Chemical Reactors because they need to have a Cupronickel Heating Coil on their bottom and a regular casing on top.
This PR would make it so that the Cupronickel Heating Coil can be in any space next to the center, not just directly below it.


Antillar:
https://github.com/Blood-Asp/GT5-Unofficial/issues/1277

PR #1300, opened 2017-11-24
Single block pump refactor & bugfix

mitchej123:
Was having trouble with pumps downstream in the GTNH fork, figured I'd offer it up here as well if there's any interest.

1) Pump will properly save it's state and reload with chunk load/reload.
It remembers that it was pumping and doesn't immediately move down.
Additionally, corrected saving of fluid type so it will properly look it
up when it reloads

2) Fixed getYHeadOfPump() so it doesn't punch through bedrock if you
destroy the pump head.  If you break a gap into an existing pipe line,
it will move down through the broken spots until it finds an existing
set of pipes - in which case it will continue from the bottom.  If you
break the pump head, it will replace the last pipe with a new Head and
carry on about it's business.

### DIFF
--- a/src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Pump.java
+++ b/src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Pump.java
@@ -1,5 +1,7 @@
 package gregtech.common.tileentities.machines.basic;
 
+import cpw.mods.fml.common.registry.GameRegistry;
+
 import gregtech.api.enums.Textures;
 import gregtech.api.gui.GT_Container_BasicTank;
 import gregtech.api.gui.GT_GUIContainer_BasicTank;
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Pump.java
```diff
@@ -122,212 +154,314 @@ public void onPostTick(IGregTechTileEntity aBaseMetaTileEntity, long aTick) {
...
-        if ((this.mInventory[0] == null) || (this.mInventory[0].stackSize < 1)
-                || (!GT_Utility.areStacksEqual(this.mInventory[0], MINING_PIPE))) {
+        if ((this.mInventory[0] == null) || (this.mInventory[0].stackSize < 1) || (!GT_Utility.areStacksEqual(this.mInventory[0], MINING_PIPE))) {
+            // No mining pipes
             return false;
         }
+
         int yHead = getYOfPumpHead();
-        if (yHead <= 0) {
+        if (yHead <= 1) {
```
draknyte1, line 274: This assumes flat bedrock.

draknyte1 (approved):
I rather appreciate the verbose comments throughout.
At the y == 1 comment I left, maybe also check if it's bedrock?

src/main/java/gregtech/common/tileentities/machines/basic/GT_MetaTileEntity_Pump.java
mitchej123: It does, but this function shouldn't otherwise punch through anything solid (the consume fluid prevents that).  

This was one of the first places I changed while tracking down the bug that was punching through bedrock -- which turned out to actually be in getYHeadOfPump() which was indiscriminately replacing blocks.

Issue #1302, opened 2017-11-26
Small bronze and steel boilers ash


SteelGiant87:
1. Currently, small boilers will stop producing ash output if the fuel type is changed
e.g. Burn some coal until one dark ash is produced, then switch to burning charcoal. The burning charcoal will never produce byproduct and the output stack will stay at one dark ash.

2. Additionally, in the code there doesn't seem to be anything that prevents the boiler from working once the byproduct stack is full, it will just keep working but stop adding byproduct (I may be wrong on this, have not been able to test).

Proposed fixes are: 

1) to change the output type to (light) ash whenever the boiler attempts to mix ash types. Light ash is chosen as it is the less valuable one, thus preventing any benefits/exploits from this change. 

2) Have the boilers explicitly check that the output is not choked before attempting to burn more fuel.

Fix 2 could be a balance change, as it would make boilers harder to automate, so this part could be omitted. I thought this was the way it was supposed to behave so I am happy to be corrected if this is not the case.

I have a pull request that fixes both these issues, and I have been able to test it.

PR #1303, opened 2017-11-26
Bugfix: Small boiler ash types mixing

SteelGiant87:
Issue: Small boilers that had ash in the byproduct slot would not produce more byproduct if the fuel type was changed to one that produced a different type of ash e.g. burn a couple of coal until there is one DarkAsh in the byproduct slot, then charcoal will produce no Ash in that stack
Fix: When mixing ash types, convert all the ash to less valuable light ashes - prevents any exploits for creating more dark ashes.
Issue: Boilers did not stop burning fuel when full of ash
Fix: added explicit check so that boilers can't burn fuel when output is stuffed (this may be a balance change, so can be removed if undesirable) 
Small refactoring of fuel logic so that ash conversion logic does not need to be repeated
Started with small bronze boiler, will extend fix to small steel boiler if general structure approved

Addresses issue #1302

### DIFF
--- a/src/main/java/gregtech/common/tileentities/boilers/GT_MetaTileEntity_Boiler_Bronze.java
+++ b/src/main/java/gregtech/common/tileentities/boilers/GT_MetaTileEntity_Boiler_Bronze.java
@@ -10,6 +10,7 @@ import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.objects.GT_RenderedTexture;
 import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_OreDictUnificator;
+import gregtech.api.util.GT_Utility;
 import gregtech.common.GT_Pollution;
 import gregtech.common.gui.GT_Container_Boiler;
 import gregtech.common.gui.GT_GUIContainer_Boiler;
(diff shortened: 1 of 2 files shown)


SteelGiant87:
I have now been able to build GT5u and test these changes. Both changes work in that small bronze boilers stop when blocked with ash, and mixed ash stacks combine to light ash.

If these changes are wanted, then I can extend them to the small steel HP boiler as well.

draknyte1:
If you're doing it for one, then it's sort of expected to change the behaviours for them all..

SteelGiant87:
Agreed, I was just starting out with one of them as the code is very similar and I didn't want to change too much duplicated code in case people wanted a different overall direction.

If you and the other devs are happy with the general direction of these changes, then I'll change the Small HP Steel boiler as well this evening.

SteelGiant87:
This change now affects both the small bronze boiler and the small steel boiler, and I have checked that it functions for both.

draknyte1 (approved):
This is ready to go.

Blood-Asp:
Another question: What happens if dust output is full?

SteelGiant87:
If dust output is full, then the boiler is unable to start burning another piece of fuel and will start to cool down.

PR #1304, opened 2017-11-26
Magneticraft & FMP Dependency

draknyte1:
Moved the 3 different API calls to be reflective, which means Magneticraft is no longer required for Dev.


Blood-Asp:
I guess the "Minor Bug" is still there?

draknyte1:
It's tagged as such because this resolves a minor bug? (Maven/Gradle not grabbing sub-deps)
And, it is also an enhancement.

Blood-Asp:
well, the "bug" tag was confusing me. i would not use it there...

PR #1307, opened 2017-11-28
Electrolysis of Sulfur Dioxide/Carbon Dioxide

JohannesGaessler:
Currently the EU cost of electrolyzing Carbon Dioxide or Sulfur Dioxide is at 14,400EU.
In consequence electrolyzing Sulfur Dioxide and then burning the Sulfur yields net energy.
While Carbon Dust cannot be directly burned you would save energy by producing it from Carbon Dioxide and not producing it from Charcoal.

This PR increases the energy cost of electrolyzing Sulfur Dioxide/Carbon Dioxide to 36,000EU.
~~In return those recipes now have a higher yield:~~
~~1000L XO2 -> 1 X Dust + 2000L O~~

~~A consequence of this change would be that you could electrolyze the Carbon Dioxide gained from some EBF recipes and then recycle the Carbon for those recipes.~~
With a MAX Voltage Muffler Hatch this process would be ~~lossless~~ 33% efficient.
This would not be realistic as in that you can't actually electrolyze carbon dioxide/sulfur dioxide but as far as I'm concerned the GT Electrolyzer has the precedent of being kind of magical anyways.


draknyte1 (changes requested):
These need rebalancing.

PR #1309, opened 2017-11-29
Fix wrong translation

Antifluxfield:


draknyte1 (commented):
What's 161 if not the correct entry?

Antifluxfield:
@draknyte1 https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/api/util/GT_LanguageManager.java#L269
“ secs”...

draknyte1:
I do wonder how that was done wrong..

draknyte1 (approved):

PR #1311, opened 2017-12-01
Fixed RawRubber from Isoprene recipe

JohannesGaessler:
Two recipes for chemically producing Rubber currently do not work correctly.
It seems I accidentally adjusted the same value twice instead of adjusting two values once each.
In consequence the recipes produced too little rubber and duped cells.


draknyte1 (approved):
Approved provided these don't need rebalancing.

PR #1316, opened 2017-12-03
Small lava boiler "calcification" mechanic

SteelGiant87:
This patch introduces a new mechanic to the small steel lava boiler, to bring it's performance more in line with the other early game single block boilers.

The boiler will produce on average one impure stone dust per bucket of lava consumed. The stone dust is placed in the output slot. Empty lava buckets are now placed back into the input slot.

The efficiency of the boiler will drop from 100% when the output slot is empty, to 33% when the output slot is full, representing the boiler becoming less efficient as it become clogged with cooled lava residues. The slot can be emptied by hand, and does not require the boiler to be break/replaced.

The reason for this change is that currently the small lava boiler is very powerful early game, and very easy to automate (a single fluid conduit). A single drum of lava from the nether takes about 10 minutes to acquire using a bucket, and generates ~25M EU in this boiler, and requires no further intervention, making lava power both least effort and the most powerful early game option.

Open to balance suggestions regarding minimal efficiency percentage and rate of efficiency loss. Currently it will take about 8 hours continuous use to reach the minimum 33% efficiency, so this represents a very minor nerf to the early game, but does make the single block lava boiler less viable to spam into the midgame.

### DIFF
--- a/src/main/java/gregtech/GT_Mod.java
+++ b/src/main/java/gregtech/GT_Mod.java
@@ -227,6 +227,7 @@ public class GT_Mod implements IGT_Mod {
         gregtechproxy.mIncreaseDungeonLoot = tMainConfig.get(aTextGeneral, "IncreaseDungeonLoot", true).getBoolean(true);
         gregtechproxy.mAxeWhenAdventure = tMainConfig.get(aTextGeneral, "AdventureModeStartingAxe", true).getBoolean(true);
         gregtechproxy.mHardcoreCables = tMainConfig.get(aTextGeneral, "HardCoreCableLoss", false).getBoolean(false);
+        gregtechproxy.mSmallLavaBoilerEfficiencyLoss = tMainConfig.get(aTextGeneral, "SmallLavaBoilerEfficiencyLoss", true).getBoolean(true);
         gregtechproxy.mSurvivalIntoAdventure = tMainConfig.get(aTextGeneral, "forceAdventureMode", false).getBoolean(false);
         gregtechproxy.mHungerEffect = tMainConfig.get(aTextGeneral, "AFK_Hunger", false).getBoolean(false);
         gregtechproxy.mHardRock = tMainConfig.get(aTextGeneral, "harderstone", false).getBoolean(false);
--- a/src/main/java/gregtech/common/GT_Proxy.java
+++ b/src/main/java/gregtech/common/GT_Proxy.java
@@ -132,6 +132,7 @@ public abstract class GT_Proxy implements IGT_Mod, IGuiHandler, IFuelHandler {
     private final DateFormat mDateFormat = DateFormat.getInstance();
     public ArrayList<String> mBufferedPlayerActivity = new ArrayList();
     public boolean mHardcoreCables = false;
+    public boolean mSmallLavaBoilerEfficiencyLoss = true;
     public boolean mDisableVanillaOres = true;
     public boolean mNerfStorageBlocks = true;
     public boolean mHardMachineCasings = true;
--- a/src/main/java/gregtech/common/tileentities/boilers/GT_MetaTileEntity_Boiler_Lava.java
+++ b/src/main/java/gregtech/common/tileentities/boilers/GT_MetaTileEntity_Boiler_Lava.java
@@ -10,10 +10,12 @@ import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.objects.GT_RenderedTexture;
 import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_OreDictUnificator;
+import gregtech.api.util.GT_Utility;
 import gregtech.common.GT_Pollution;
 import gregtech.common.gui.GT_Container_Boiler;
 import gregtech.common.gui.GT_GUIContainer_Boiler;
 import net.minecraft.entity.player.InventoryPlayer;
+import net.minecraft.item.ItemStack;
 import net.minecraftforge.common.util.ForgeDirection;
 import net.minecraftforge.fluids.FluidStack;
 import net.minecraftforge.fluids.IFluidHandler;
@@ -24,6 +26,8 @@ public class GT_MetaTileEntity_Boiler_Lava
         super(aID, aName, aNameRegional, new String[]{
                 "A Boiler running off Lava",
                 "Produces 600L of Steam per second",
+                "Drops to 200L of Steam per second as byproduct fills",
+                "Clean out byproducts to keep efficiency high",
                 "Causes 20 Pollution per second"});
     }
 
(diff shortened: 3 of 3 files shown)


draknyte1:
Make it configurable and default off. 

Lava boilers require steel, nether requires steel, it's not really exploitable. 

If it's not automatable, I'd give this less weight as a wanted feature. (People don't like losing automation early game)

Don't balance features based on pipes from EIO.

draknyte1 (changes requested):
As above.

SteelGiant87:
I'll try to get back to this later in the week.

SapientMC:
>  A single drum of lava

Just don't put ExUtilities to modpack with GT.

SteelGiant87:
> Just don't put ExUtilities to modpack with GT.

This change will have no effect on standalone GT early game if players are still hand loading the machines with buckets, as then they will still be able to do that and clean out the slot at the same time. I was pointing out one way in which this block is much more powerful than the other single block boilers of its tier, which allows it to massively outperform them in the early and mid game when used in the most popular modpacks.

The objective is to bring this machine in line with the other early game single block boilers, which have all been modified (not by me, [by Blood Asp himself](https://github.com/Blood-Asp/GT5-Unofficial/commit/873902f4bf6c97a9ab33c18cce62a7cdfebfd9c0#diff-89495a35f7036362312fb25f74072dc0)) to be impossible to automate: the small boilers by not being able to pipe (char)coal in, and the small solar boiler by the calcification mechanic requiring the player to break and replace periodically. This change does not make it impossible to automate the boiler, unlike the other options, it merely slightly lowers the efficiency over time (and never to zero) so that while this block is _still_ both the most powerful and the easiest to use of the single block boilers, it has a minor drawback that will make the user want to move onward to higher technologies.

SapientMC:
Ok. I agree. High Pressure Coal Boilers can not be fully automated and will stop working when output slot are full. Think HP Lava Boiler should do the same instead of working with less efficiency.

SteelGiant87:
I might be able to get back to this tomorrow, but I now have two conflicting suggestions: one to make this change optional, and one to make it significantly harsher and stop working entirely.

I'm not sure who to follow. Is there even a way to get a list of who has access to a repository?

SapientMC:
Сalcification and efficiency mechanic are made for "Low pressure" Boilers. HP Boilers works with full efficiency until they can. HP Coal already has the conditions for stopping work, HP Lava should have it too.

draknyte1:
There's all of three (Blood, Johannes & Me) of us who have made a commit in the last 6months. 

I meant removal of output automation makes the change shit. If manual clearing of the output slot is required, its a horrible change. 

I'm not sure how it's so hard to do both. 
Make it a config option, make it lose efficiency and make it lose efficiency until full. I don't think any boilers should stop when full, because they require so much manual maint. The key point being raised is a thing called consistency. All small boilers should have the same mechanics. The solar boiler really being an exception.

leagris:
I advocates not doing these changes.
What real issue are you trying to fix?
Do you have citations of real cases where players actually abused/spamed small boilers or broke game balance in some way using these small boilers?
Does you change provide a better game-play with new features or help players evolve with better power sources?

I think players are already doing too much nursing of these boilers with manual feeding fuel they probably also manually harvest from a charcoal pile, or manually mine from coal/lignite veins doing repeated back and forth travels between their base and fuel source; without any automation available within Gregtech, at such low level tech where small boilers are a must have; because these are the sole option besides spaming already handicapped solar boilers.

I think you are trying to make player experience even worse.
You all already turned Gregtech5 into a perpetual frustration. This is the kind of change that add even more restrictions instead of expanding options and creativity.

For this reason I disapprove this change or stopping boilers once output slot is full of ashes.

NNM42:
agree with Lea

SteelGiant87:
> There's all of three (Blood, Johannes & Me) of us who have made a commit in the last 6months.

Really? I asked because I saw about 10 names just looking through the last couple of months of commits.

> I meant removal of output automation makes the change shit. If manual clearing of the output slot is required, its a horrible change.

I was suggesting that as a mechanic, as it would be consistent with the small bronze/steel coal boilers, which were explicitly modified to be unable to automate cleaning of output. I can try and make it possible to remove stuff from the output slot automatically, although I note that the base single block class was changed to explicitly disable this.

> I'm not sure how it's so hard to do both.
> Make it a config option, make it lose efficiency and make it lose efficiency until full. I don't think any boilers should stop when full, because they require so much manual maint. The key point being raised is a thing called consistency. All small boilers should have the same mechanics. The solar boiler really being an exception.

Ok, I can make it a config option. The current implementation does as you ask and makes it lose efficiency as output fills and it does not stop when full, just bottoms out at 1/3 production rate. Let me know if the current implementation fits with what you want.

SteelGiant87:
@leagris my intention is not to make gameplay a more frustrating experience, but to make climbing the tech tree a more rewarding exercise.

I remember when a couple of years ago I set up a nuclear reactor, costing thousands of metal plates of various varieties for the blocks and components, and a complex fuel processing chain, and it only outputs a few hundred EU/t. A reactor now seems to be gated behind EV components, although the fluid reactors look to have had a buff. (Please correct me if I'm wrong, I'm only at MV on my current playthrough).

These single block lava boilers cost 10 steel each. They produce 600 steam/second, which is 30 steam/tick. In an LV turbine this converts 2 steam to 1 EU at 0.85 efficiency, so a single boiler can produce (30 / 2) * 0.85 EU/t = 12.75 EU/t. You can get 1A of MV for 100 steel, and 1A of HV for 400 steel. The fuel for these boilers is effectively limitless as it can be obtained from a dimension that is literally full of the stuff.

In my current playthrough I had intended to use diesel generators, however when I saw how powerful these boilers were, and the majority of my effort was spent gathering and refining oil to fuel, the fact that I could have instead just set down one pump anywhere in the nether and had a better return in vastly less time made oil look effectively useless as a power source in comparison. This was my motivation for trying to find a change that doesn't nerf lava boilers into uselessness, but also makes them not all of the cheapest and easiest and most powerful to use.

These features combined makes the lava boilers look out of place in the progression to me. I took from the changes to the other low tier single block boilers that the design intent in modifying them was to make it less desirable to spam many low single tier boilers by preventing automation, in order to make the more mechanically rich multi block boilers a better choice once they became available.

As I see it, this change would not significantly degrade the capabilities of the lava boiler within the small number of hours it takes to obtain other power sources, thus not significantly increasing grind in the early game, but does encourage players to move on to new options once they have the capability.

draknyte1 (changes requested):
Still requires a config option.

Blood-Asp:
The calcification of the Solar Boiler has also no config option. But it is missing a way to show the current state in the gui and clean it without pickup+place. So i would really like to see these functions added.

All the steam age boilers have some not automatable function that will make them stop working or low efficency after a few days without maintainance. Only the lava boiler was missed by me.

The efficency of the lava boiler is really high, even the large heat exchanger can only top it with superheated steam. And yes, i saw many that were spamming them mid game for HV+ power.

And as it was mentioned above: The Nuclear was boosted a lot. Default EU direct output 2xEU/t. Fluid reactor 4x boost. using fluid reactor + Large heat exchanger + superheated steam i managed to get 8800EU/t from 4 quad fuel rods. Before the reactors where about HV tier and at the same level as many other energy sources. now it fills the hole beween the HV powergens and fusion.

SteelGiant87:
Apologies for not being able to get back to this, been busy in the run up to Christmas, hopefully have time again soon.

I would be keen to do as you say and add some UI information and non break/replace cleaning of the solar boiler.

Are we now leaning towards neither or both solar and lava boilers having efficiency configs?

Really looking forward to trying out the new (for me) nuclear.

Blood-Asp:
@SteelGiant87 I'm now back on merging/Fixing at last for a few days and Christmals is also long over. How is it looking for you? Going to finish this up?

SteelGiant87:
Apologies, more things came up.

I will try to get it done by this weekend.

SteelGiant87:
Ok, finally had a chance to put this in. There is now a config option that allows this feature to be disabled. If the option is disabled, the byproduct will still be created, but there will be no efficiency decrease.

(Also a minor enhancement: when efficiency is lower, water consumption is reduced proportional to the reduction in steam output.)

draknyte1 (approved):

PR #1318, opened 2017-12-07
Fixed the spelling of Colour across the entire project.

draknyte1:
Because we're not American.
If we were, we'd also be using Aluminum, but we don't.


JohannesGaessler:
I generally don't care which spelling standard we use (except for Aluminium).
However, I think that if we're gonna choose one we should document it somewhere, preferably in the file that sets standards for PRs.
Also, what about other words that follow the American spelling conventions, such as 'centered' instead of 'centred'?

src/main/java/gregtech/api/gui/GT_GUIContainerMetaTile_Machine.java
```diff
@@ -27,12 +27,12 @@ public GT_GUIContainerMetaTile_Machine(InventoryPlayer aInventoryPlayer, IGregTe
...
-            if (tColor < ItemDye.field_150922_c.length) {
-                tColor = ItemDye.field_150922_c[tColor];
-                GL11.glColor4f(((tColor >> 16) & 255) / 255.0F, ((tColor >> 8) & 255) / 255.0F, (tColor & 255) / 255.0F, 1.0F);
-            } else GL11.glColor4f(1.0F, 1.0F, 1.0F, 1.0F);
-        } else GL11.glColor4f(1.0F, 1.0F, 1.0F, 1.0F);
+        if (GregTech_API.sColouredGUI && mContainer != null && mContainer.mTileEntity != null) {
+            int tColour = mContainer.mTileEntity.getColourization() & 15;
+            if (tColour < ItemDye.field_150922_c.length) {
+                tColour = ItemDye.field_150922_c[tColour];
+                GL11.glColour4f(((tColour >> 16) & 255) / 255.0F, ((tColour >> 8) & 255) / 255.0F, (tColour & 255) / 255.0F, 1.0F);
```
Quantum64: You can't change the spelling on GL11 methods. This won't even compile.

draknyte1:
Any GL functions were fixed in commit 2. 
https://github.com/Blood-Asp/GT5-Unofficial/commit/259ca719445a6151d3e2ddb7f2fd041c313c5985

It has 0 project errors and should compile fine.

Quantum64:
I see

Antifluxfield:
Wait... What about the backward compatibility? Like the tiles colored by tag "mColor", and the addons referencing those interfaces, methods and fields (if there are any of them)?

Blood-Asp:
this seems useless to me

PR #1319, opened 2017-12-07
Fixed Mufflers and Large Boiler Fuels.

draknyte1:
Mufflers can no longer face downwards, into a machine.
This abuse allowed machines to check inside themselves for air.

Large Boilers cannot take Lava anymore, as originally intended.

### DIFF
--- a/src/main/java/gregtech/GT_Mod.java
+++ b/src/main/java/gregtech/GT_Mod.java
@@ -1198,7 +1198,6 @@ public class GT_Mod implements IGT_Mod {
         		GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Sulfur, 1),
         		GT_OreDictUnificator.get(ItemList.Block_SSFUEL.get(1, new Object[0])),
         		GT_OreDictUnificator.get(ItemList.Block_MSSFUEL.get(1, new Object[0])),
-        		GT_OreDictUnificator.get(OrePrefixes.bucket, Materials.Lava, 1),
         		GT_OreDictUnificator.get(OrePrefixes.rod, Materials.Blaze, 1));
         if (Loader.isModLoaded("Thaumcraft")) {
         	GT_Recipe.GT_Recipe_Map.sLargeBoilerFakeFuels.addSolidRecipe(GT_ModHandler.getModItem("Thaumcraft", "ItemResource", 1));
--- a/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_Muffler.java
+++ b/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaTileEntity_Hatch_Muffler.java
@@ -54,8 +54,8 @@ public class GT_MetaTileEntity_Hatch_Muffler extends GT_MetaTileEntity_Hatch {
 
     @Override
     public boolean isFacingValid(byte aFacing) {
-        return true;
-    }
+        return aFacing != 0;
+    }    
 
     @Override
     public boolean isAccessAllowed(EntityPlayer aPlayer) {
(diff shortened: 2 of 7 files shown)


MauveCloud:
> Mufflers can no longer face downwards, into a machine.
This abuse allowed machines to check inside themselves for air.

Hang on.  That would only be checking inside the machine for machines that require the muffler to be at the top above a hollow section.  A large boiler requires the muffler hatch in the bottom layer, at which point having it face down toward an air block would not be checking inside the machine, and should be allowed.  A large gas/steam turbine or large plasma generator allows the muffler hatch to be on the top, side, or bottom.  If placed at the bottom, facing downward would be okay, but facing upward should be forbidden, and if placed at the side, determining which way is toward the inside of the machine would be a little trickier.

draknyte1:
No machine REQUIRES it facing downwards. Which is far more exploitable than any other direction.

leagris:
@draknyte1 Multimachines are cuboid.
Should be easy enough to check the air block is not within the cuboid structure boundaries instead of hardcoding a directional check.

MauveCloud:
That's not universally true, but the multiblock machines that aren't cuboids don't use muffler hatches, so that seems like it will work.  However, draknyte1's proposed solution seems to change whether a muffler hatch is even place-able/wrench-able to face downward, whereas checking the cuboid would have to be added to the multiblock controller (since the muffler hatch by itself doesn't know where the structure boundaries are).

leagris:
Then having the Controller perform muffler hatch orientation check during check machine and report incomplete structure if muffler is in wrong orientation or has output obstructed could be a valid alternative.
Afterall, the air block beside a muffler output is part of the multiblock structure.

draknyte1:
Yeah, or, just leave the simple fix as it is. 

Maint. Hatches can only face horizontal, it's never caused an issue. This won't either because there is zero legitimate use for facing them down.

SapientMC:
@draknyte1 
I disagree with 
> zero legitimate use for facing them down.




Middle turbine consume two different fuel type with its own optimal flow, so both top center blocks are occupied by input hatches.

draknyte1:
I'd rather change where hatches go on turbines than change the current status of the PR.
95% of cases where they face down are people exploiting them to face internal.

I'll fix that tomorrow or something.

SapientMC:
The turbine has only 2 blocks of air inside. It's logical to check if the muffler is directed to one of them and that's it.

OvermindDL1:
> 95% of cases where they face down are people exploiting them to face internal.

Not really, I placed them facing down underneath my machines in the majority of cases.

(I didn't actually think of facing them internally, I figured that'd already be checked for...)

Blood-Asp:
I somehow mean to remember that back in GT4 muffler hatches only could be placed on the top side of turbines but not sure. And as i tought so i never checked if they can work on other sides. But facing down for an exhaust actually feels wrong to me... 

Also about the comments to check if they face inside: The way they are currently implemented that would not be so easy. Actually can't thing about an easy way right now at all.

draknyte1:
There is no easy way, simply put.
Mufflers facing down is invalid and I agree with your statement on the fact it feels wrong.
But hey, maybe Pollution is denser than air and falls downwards?

I left this as In Progress because I wanted to think of an alternative solution over the next few days.

SapientMC:
I want to remind you that the exhaust pipe of all cars is located below.

draknyte1:
And they face horizontally in all 16 countries I've visited. 
Your point is invalid.

SapientMC:
But none of them occupies a square meter of surface.

leagris:
I am just sarcastic but:
Gregtech is known to place multiblock devices the most annoyingly possible.
So if it is convenient to have a muffler hatch at bottom of turbine, be sure it will be denied because Gregtech is not convenient.
Why do you think turbine Controller is at the most annoying place possible? 😆

OvermindDL1:
> But facing down for an exhaust actually feels wrong to me...

@Blood-Asp Well at work the emergency generator exhaust faces down to prevent rain from getting in to it.

And yes, pollution is denser than air.

Issue #1293, opened 2017-11-19
Phosphorous Pentoxide material waste


Antillar:
Currently Phosphorous Pentoxide is crafted this way:
4 P + 10 O -> 1 P4O10
Should be
4 P + 10 O -> 14 P4O10
So no materials are wasted

draknyte1:
@JohannesGaessler Can't balance equations and as far as he is concerned 1 P4O10 is worth 4P and 10O, which is wrong.

Antillar:
https://github.com/Blood-Asp/GT5-Unofficial/pull/1322

Issue #1292, opened 2017-11-19
Sodium Sulfide material waste


Antillar:
The current recipe for Sodium Sulfide is:
2 Na + 1 S -> 1 Na2S
It needs to be
2 Na + 1 S -> 3 Na2S
So no materials are wasted.

leagris:
2 Na + S → Na₂S

2 × dustSodium (oredict) + 1 × dustSulfur (oredict) → 3 × SodiumSulfur dust (items)
KISS preferred in a game like Minecraft: 2 Na + S → Na₂S

Although in real life: this chemical reaction is a laboratory and costly one that needs a THF catalyst (Tetrahydrofurane or 1,4-Epoxybutane or Oxolane or Oxacyclopentane or Tetramethylene-oxyde).

The cheapest industrial processes for Na₂S production are:
- Sodium sulfate and Carbon: Na₂SO₄ + 4 C → Na₂S + 4 CO
- Sodium sulfhydrate and Soda: NaSH + NaOH → Na₂S + H₂O

Antillar:
@leagris I take into consideration the backward recipe. From 3 Na2S in Electrolyzer the player can obtain 2 Na and 1 S. 
I do like the concept you suggested more, but currently GregTech follows the "lossless" pattern of chemical recipes, and there are countless examples to this. And this is an exception I want to be removed.

Techlone:
@Antillar it's a wrong way. @JohannesGaessler brought new chem and he has followed real chemical rules.
I hope in someday all electrolyzer recipes be correct so you will get 2Na and 1S from 1Na2S.

Antillar:
https://github.com/Blood-Asp/GT5-Unofficial/pull/1322

PR #1322, opened 2017-12-08
High density Electrolysis/Centrifugation recipes

JohannesGaessler:
The materials I've been adding which are used in chemical recipes generally have a density equal to the size of their formula.
Currently their reverse recipes (Electrolysis/Centrifugation) however still work according to them having a density of 1.
In consequence there is a net loss of material whenever you revert one of those materials back into elementary materials.

Edit: When I say density I do not mean physical density. I mean the density in terms of material units in one unit of a compound material.

When I first added those materials I was aware of this problem but did not yet implement a solution because I wanted some time to think about it first.
I ultimately settled on the continued use of the already existing machines but at an increased EU cost.
Specifically the EU cost is calculated as follows:

* 18,000 EU per Hydrogen, because otherwise electrolyzing hydrocarbons would be too good of a source of Hydrogen
* 24,000 EU per Oxygen, because Materials containing Oxygen generally have a reduced burn value
* 12,000 EU per any other Material

New recipes run at 120 EU/t for Electrolysis, 30 EU/t for Centrifugation.
Recipes for old materials remain unchanged.
I think the energy costs for the new recipes might be too high, feedback is appreciated.

This PR adjusts the Electrolysis of Water to make it consistent with the new chemical usage (recipe now consumes 1000L instead of 3000L).
On an unrelated note I fixed a bug where the wrong amount of Quicklime was used in one of the EBF recipes for increased Iron output.

I made the following changes to the API:

* Materials now have a boolean 'reverseRecipeEUByMaterialValue' to determine which method should be used to calculate reverse recipe EU cost. By default the old method is used.
* When adding a reverse recipe the amount of inputs and outputs are now reduced by the greatest common divisor before the recipe is added. This reduces the overflow of ItemStacks, i.e. you don't get a distribution where you have several stacks of the same material with a combined stack size < 64.


Antillar:
https://github.com/Blood-Asp/GT5-Unofficial/issues/1293
https://github.com/Blood-Asp/GT5-Unofficial/issues/1292

draknyte1 (changes requested):
Don't rebalance chemistry wrong. 
I can provide about 10+ people from various modpcks that have unanimously agreed that your way is wrong and the way chemistry is done in the rest of Gregtech is correct. 

1000L of Oxygen + 2000L of Hydrogen = 3000L of water. Your 5% of the chemical recipes the currently exist are all wrong and you'd be better to rebalance them to be consistent with the REST of the mod.

JohannesGaessler:
@draknyte1 
If the prevalent opinion was that the way I balanced things is wrong then I would have gotten immediate feedback regarding this when I made the original PR, not six months after it has been merged.

leagris:
@JohannesGaessler I did not test 5.09.31 until lately, so I had not realized the impact of your chemistry changes.

I confirm that: while your chemical balancing may be more realistic; it does not fit right with everything-else in Gregtech or even other mods.

Please reconsider this quantities balance and do it constant volume.
Even if it looses on realism:
It is the way it has been handled previously for good reasons:
I am pretty sure it fits better with a fantasy universe, that just happens to have similarities with our real World, but is fundamentally very different, and 10^∞ less complex.

NNM42:
/0\ WTF?

draknyte1:
@JohannesGaessler, after a while people have come to realise that your way is incorrect. 
Mistakes aren't always noticed immediately. 
Look at the Apollo space missions. 

Current input from anyone consulted has said that 2000+1000=3000, not 1000.

SapientMC:
2 moles of hydrogen + 1 mole of oxygen = 1 mole of water

draknyte1:
For the sake of consistency, and communal opinion, 2+1=3, because we're not measuring in moles.

leagris:
There is only one device that account for molecular mass: Matter Fabricator
And that's it all. It is a good helper indicator to balance replication cost of different materials.
It is not appropriate for all chemistry because it a simplified and tuned rules chemistry based on constant volumes, pulling aside any consideration of mass or mole.

SapientMC:
@draknyte1 The amount of the substance and its volume (in which we measure) are completely correlated.

SapientMC:
@leagris 
> It is not appropriate for all chemistry because it a simplified and tuned rules chemistry based on constant volumes.

Why it should stay simplified for ages?

draknyte1:
So @SapientMC, if they correlate, then 2000l of Hydrogen + 1000L of oxygen makes 3000L of water. 
That is Correlation, which you so strongly suggests exists. (It does)

Unless you redo all Chemistry, than Johannes way of doing things is inconsistent and wrong.

SapientMC:
IRL at least in Russia, chemistry is beginning to be studied in the 8th grade of the school (about 14 years). So I see no reason to leave it as simple as possible.

draknyte1:
Good for Russia. 
Chemistry is optional in the rest of the world. 
We are not making life simulator, it doesn't have to be accurate. 
We are making a large MC mod that REQUIRES consistency. 

Which is what Johannes has not achieved. 

So unless you redo ALL of the chemistry to be consistent, it's not worth having 5% of recipes be totally fucking different.

SapientMC:
@draknyte1 
> then 2000l of Hydrogen + 1000L of oxygen makes 3000L of water.

No way! 2kg of hydrogen + 1 kg of oxygen  makes 3kg of water! 
During chemical reactions, the mass is retained and not the volume

Rongmario:
If you're changing that, change the entire mod.

We're talking consistency here.

Antillar:
Density of Water: 1000 kg/m3, 3000 kg of water is 3000 L.
Density of Hydrogen Gas: 0.09 kg/m3, 2000 kg of this gas is 22 222 222 L.
Density of Oxygen Gas: 1.470 kg/m3, 1000 kg of this gas is 680 272 L.

**1 mb of Water gives 7 403mb of Hydrogen and 226 mb of Oxygen.**

Want realism?
THIS is realism. The rest is bullshit!

JohannesGaessler:
@Antillar When I say density I do not mean physical density. I mean the density in terms of material units in one unit of a compound material

Antillar:
@JohannesGaessler the name for what you mean is Molar Mass or Molecular Mass

Antillar:
@JohannesGaessler currently in GT 1 bucket equals 1 mole of smth. There is 1 mole of helium plasma obtained from hydrogen isotopes. There is 1 mole of chlorine obtained from salt.
However, this only applies provided that none of the materials are metals, then 1 mole is 1 dust, or 0.114 bucket, and 1 bucket of molten iron f.e. contains several moles of iron.
Let the water electrolization be the way it fits into GT: 3 moles of water into 2 moles of hydrogen and 1 mole of oxygen.
This is NOT realistic by any means. But it fits in current system.

Antifluxfield:
@Antillar That's not a mole of something... What we applied is a strange "atomic" measurement: each Liter of fluid means a unit of various atoms, not a unit of molecules, and the latter one is just what the "amount of substance" means. Thus we have 2000L Hydrogen + 1000L Oxygen = 2000u Hydrogen atoms + 1000u Oxygen atoms = 3000u (various) atoms = 3000L water.

PR #1334, opened 2017-12-15
Chemistry revision: Volume conservation and simpler recipes

JohannesGaessler:
This PR rebalances chemical recipes to conserve volumes, adds simple variants of chemical recipes, and revises cracking to make it more ergonomic.

# Volume Conservation

When I originally added the overhauled chemical recipes I broke with the prevalent convention of conserving material volumes.
I did this because it resulted in recipes with a better ratio of inputs/outputs and I judged this to outweigh the introduced inconsistency.
This judgement has turned out to be wrong.
This PR rebalances the inputs/outputs of the new recipes to generally conserve volumes.

# Simple recipes

When I added the recipes I added them in a way that I would enjoy them as a player.
I enjoy complicated recipes with equally complicated logistics, but I neglected that a large part of the playerbase does not enjoy this level of complication.
As such, this PR adds simpler variants of the chemical recipes I implemented.
Whether the simple or the complicated recipes are being used can be controlled via the new config "MoreComplicatedChemicalRecipes" in Gregtech.cfg under general.
By default the simple recipes are used.
The rationale behind this is that new players, who do not know about config options would in all likelihood have an easier start with the simple recipes.
Also, players who prefer the complicated recipes will likely be less averse to the simple recipes than vice versa.

# Cracking revision

When I added the current cracking recipes I added three recipes per cracked material and per cracking type (Severely Steam/Hydro-Cracked, Moderately Steam/Hydro-Cracked, Lightly Steam/Hydro-Cracked).
The idea behind this was to implement Steam-Cracking for plastic production and Hydro-Cracking for fuel refining. 
The severity parameter controls the ratio of the outputs.
However, I think the severity parameter is fundamentally not necessary.
The amount of time that you would spend figuring out how to set the severity is never worth the negligible efficiency increase.
Also, having three recipes per Cracking type makes it much more difficult to look through the cracking recipes and to understand what they do.
This PR reduces the number of Cracking recipes to two per material (one recipe for Steam-Cracking, one for Hydro-Cracking).

Also, Cracking now only becomes necessary at EV in conjunction with a Distillation Tower.
Intermediary Oil products like Ethylene can now be directly distilled from Naphtha.

These changes apply to both the simple and the complicated recipe setting.


Antillar:
Well, it is the time to feel kinda tricked or something. I have spent many hours to completely understand the oil processing system, I have calculated all values, solved the recursive set of equations of the values, calculated the overall requirement in every type of plastic and percent requirement. I have solved the set of equations which gave me a clear amount of every hydrocarbon required, which substituted the respective variable in the first recursive set, and then I solved it too. 
The perfect cracking options are:
Light on RG, Light on HF, Medium on Naphta, Severe on LF, Light on Propane, Medium on Ethane, Light on Butene. 

This produces all required materials in the proportion they are used by an average player. The SBR is not produced in favor of Si Rubber, as more efficient advanced rubber. However, I have decided to not to crack butadiene, to save it for LuV+.

The total amount of crackers: 7
Distillation towers: 8
LCRs: 8
This was done on .31 with 1000L of Acetone left on every 1000L of Epoxy produced. I have used it to produce Solid Super Fuel, which I already have in the amount of over 3k blocks.

JohannesGaessler:
I'm sorry for having baited you into wasting a lot of time and effort.
Do you think the more complicated Cracking system should be kept?

Antillar:
No. I am wiki contributor, so it is my job to fully understand every mod mechanic of the mod I am documenting. Also I have an LP world with everything done, so I was able to oversee my progression and calculate requirement proportion. And I have a magister degree in mathematics, so I was able to solve all this.

Normal player has neither of those. So I am absolutely conserned that there is no reason to return to previous cracking even with config option. I have not heard of anyone else figuring the proportions. So I think that this mechanic is too hard for an average player.

SteelGiant87:
I think the change to cracking makes a lot of sense: it retains the meaningful choices, while streamlining the gameplay significantly.

I also think the change to volume measurements, even though it isn't real chemistry, will make things better by being consistent.

Well done for having the courage to change your design and listening to feedback.

jaskarth:
We might need to keep these recipes in for GTNH's sake, as it takes the "make everything more complicated" approach.

draknyte1:
@SuperCoder7979 What GTNH keeps has no effect on master. What we do, should not consider them. 
It's literally their choice if they cherry pick the PR and merge it into their fork. 

@Antillar making config options for things is nice but having a config option for this is not. 
Literally would be a gigantic cluster fuck of additional code spread across the mod, so no config. 
Not the first recipe change, but hopefully the last major one.

draknyte1 (changes requested):
Don't leave the old content. 
It's no harder than what the revised recipes should be. Because it's all "balanced". 

There is 0 reason to leave the old recipes when we've agreed they're all wrong.

jaskarth:
Alright, I agree.

JohannesGaessler:
@draknyte1 I think there might have been a misunderstanding.
The config option does not enable chemical recipes as they currently exist with varying material densities.
I have rebalanced recipe material ratios for the simple recipes and the more complicated recipes.

draknyte1:
I still don't see why you're needing a config option..

PR #1337, opened 2017-12-18
Fix BMP texture assets with png extension

leagris:
Converted to PNG as it should be.
```Bash
fgrep -rl BM6 *.png | xargs -n1 mogrify -format png
```


draknyte1:
Does that mean you did these incorrectly the first time in https://github.com/Blood-Asp/GT5-Unofficial/pull/427/files ?

Because the difference is only 3 files on top of the previous 38 you did > a year ago.

leagris:
It means it was not merged or someone rolled-back BMP files

draknyte1:
They were rolled back at some point. 

What is the actual point of the PR though?
Like, it doesn't change anything at all as far as anyone will know, doesn't add performance, fix bugs or anything. No OS has issues decoding the images either, so it's not even a fix in that sense.

leagris:
It is fixing discrepancy between file-names and image format.
It is using more modern, better lossless compression, less space PNG format versus BMP format.

draknyte1 (approved):
Original PR was merged to master branch and not unstable.

Blood-Asp:
and this commit goes into experimental and not unstable. so also not really usable too...

leagris:
Provided a PR with same changes for the unstable branch https://github.com/Blood-Asp/GT5-Unofficial/pull/1345

To be sure all png files are proper png format and not alien BMP Bitmap, you can use this simple one-line command:
```Bash
find ./GT5-Unofficial/src/main/resources/assets/gregtech/textures/ -type f -name \*.png -exec fgrep -rl BM6 {} \; | xargs -n1 mogrify -format png
```

PR #1341, opened 2017-12-20
Custom Materials Fix

Antifluxfield:
Fix the material name of custom materials, along with the oredict name & other relating things.
Fix the NPE in init phase caused by that some materials loaded as null.


draknyte1 (approved):

Blood-Asp:
I guess these that have already added custom materials with the config now need to adapt the change.

PR #1342, opened 2017-12-21
Rework gasoline

jaskarth:
Previously I made high octane gasoline output 25000L of fluid out of the mixer, which caused bugs due to the mixer only having an output buffer of 8000L. By changing the premium gasoline to output as cells this issue has been fixed. The only major change to peoples' builds using this is the addition of a fluid canner to turn the gasoline cells into fluid for use in the diesel engine.


draknyte1 (approved):

draknyte1 (changes requested):
I can't math. 
Rebalance discussion took place on discord.

PR #1346, opened 2017-12-23
Localization

Antifluxfield:
In #1251 , I introduced the "%material" placeholder to better support localization; but seems that it [doesn't work well as what I thought in some languages](https://forum.industrial-craft.net/thread/11488-gregtech-5-1-7-10-forge-1355-unofficial-but-approved-port-stable-even-gt5-experi/?postID=211824#post211824). Though you can overwrite the placeholder if necessary, the lang file is not that readable anymore. Thus, as a simple solution, I add a config to disable this placeholder in this PR.


Antifluxfield:
The last two commits fix some other l10n issues.

src/main/java/gregtech/api/metatileentity/MetaPipeEntity.java
```diff
@@ -92,12 +96,16 @@ public MetaPipeEntity(int aID, String aBasicName, String aRegionalName, int aInv
...
-        if (GT.isClientSide()) {
-            ItemStack tStack = new ItemStack(GregTech_API.sBlockMachines, 1, aID);
-            tStack.getItem().addInformation(tStack, null, new ArrayList<String>(), true);
+        if (aAddInfo && GT.isClientSide()) {
+            addInfo(aID);
         }
     }
 
+    protected final void addInfo(int aID) {
+    	ItemStack tStack = new ItemStack(GregTech_API.sBlockMachines, 1, aID);
```
mitchej123: You should add server side protection here, otherwise if someplace calls this without the check it'll crash.

Antifluxfield: Yep

PR #1349, opened 2017-12-24
GT6 cable quality of life improvements

mitchej123:
Allow use of wire cutter/soldering item on the tile entity, as well as the wire, for a connect/disconnect. Useful for when the cable you want to power the machine is hidden behind the machine and not easily accessible.

Porting upstream from https://github.com/GTNewHorizons/GT5-Unofficial/pull/64

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaPipeEntity_Cable.java
+++ b/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaPipeEntity_Cable.java
@@ -2,6 +2,7 @@ package gregtech.api.metatileentity.implementations;
 
 import cofh.api.energy.IEnergyReceiver;
 import gregtech.GT_Mod;
+import static gregtech.api.enums.GT_Values.D1;
 import gregtech.api.GregTech_API;
 import gregtech.api.enums.Dyes;
 import gregtech.api.enums.Materials;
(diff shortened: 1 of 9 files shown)


Technus:
Just for a heads up might fuck up current behaviour of soldering iron on machine sides... 
which changes the strenght of redstone signal emmited

Technus:
By that i mean it overlaps with it, even if there is a check what should happen an overlapping behaviour is bad imo.

Just a matter of preference.

Technus:
Other idea would be to implement other tool for "electical" wire soldering/cutting. 

like electric wire cramping/cutting tool

mitchej123:
Reverted the soldering iron changes; left the wire cutter right click behavior.

Technus:
@mitchej123 cherrypick this commit to that merge branch
https://github.com/GTNewHorizons/GT5-Unofficial/commit/c89acf537d0b729f41147f8575016d96f53c6ee2

mitchej123:
Included a fix for an issue with gt6 style wires and GT solar panels (covers) that some players noticed.  CC 

CC @Antifluxfield

src/main/java/gregtech/api/metatileentity/implementations/GT_MetaPipeEntity_Cable.java
```diff
@@ -269,29 +275,61 @@ public int connect(byte aSide) {
...
+            if (tTileEntity instanceof IColoredTileEntity) {
+                if (getBaseMetaTileEntity().getColorization() >= 0) {
+                    byte tColor = ((IColoredTileEntity) tTileEntity).getColorization();
+                    if (tColor >= 0 && tColor != getBaseMetaTileEntity().getColorization()) {
+                    	return rConnect;
+                    }
+                }
+            }
+            
+            boolean sHasSolarPanel = coverBehavior instanceof GT_Cover_SolarPanel;
```
mitchej123, line 299: Mentioned downstream - I don't love this; but I'm not aware of any other "covers" that inject energy. Also it's the only connection that doesn't require a tile entity on the other side.

mitchej123:
Added a fix for cables & AE2 GT EU p2p tunnels

PR #1363, opened 2018-01-03
Oredict support for AssLine Recipe

Antifluxfield:
Now you can register assline recipes with oredict; or just with several alternative items as one of the ingredients.


Antillar:
Finally dangt

NNM42:
Antifluxfield, plz see this https://github.com/Blood-Asp/GT5-Unofficial/issues/1411

PR #1368, opened 2018-01-07
Fix input & output condition of fluid pipes

Antifluxfield:
https://github.com/GTNewHorizons/GT5-Unofficial/issues/81


mitchej123:
This will actually prevent any pipes from connecting to each other.

I have a fix that works in drain and fill; will open a PR here.

mitchej123:
https://github.com/Blood-Asp/GT5-Unofficial/pull/1371

Antifluxfield:
Ummm... I know what's wrong...
Your commit should be a proper solution.

PR #1371, opened 2018-01-09
Update drain/fill to take into account pipe connections

mitchej123:
This should stop things (output hatch, for example) from moving fluid into unconnected adjacent fluid pipes.

Additionally consolidated the logic into a helper function.


draknyte1:
Things like Fluid Output Hatches are intended to automatically output their contents, iirc.
Are you talking pipes like this?

X: Casing
O: Output Hatch
P: Pipe
--------------------
```
XX
XOPPP
  P<-- this one?
```

mitchej123:
They are currently outputting to unconnected fluid pipes; this requires a fluid pipe be connected before they can be output to.  It doesn't change the behavior of anything else in front of the fluid output hatch, or the behavior of connected pipes.

I'm not sure what your diagram is trying to illustrate.

mitchej123:
Here's some screenshots to illustrate what I changed:

This previously output (but in my mind shouldn't since the pipe isn't connected):
should_not_output


This previously output (and still does, since the pipe is connected):
should_output

PR #1374, opened 2018-01-13
Fix custom enchantment for materials

Antifluxfield:
Enchantments should be initialized and registered before loading material configs, or the customization of enchantment of some materials won't work.


draknyte1:
Please give the required details when submitting PRs or they'll be closed in the future without warning.

PR #1378, opened 2018-01-16
Fix: Fluid pipes transfer capacity

shpakovivan:
**Fixed:** The pipe lost its transfer capacity if one pipe block had more than one consumer.

**Added:** The fluid distribution algorithm is optimized. Now the fluid flows along the path of least resistance.
Not all GT-machines return fluid tank info. For these machines, a fluid pipe use a capacity similar to its own (considering it to be empty).

Illustrations attached:
mc1
mc2
mc3
mc4
mc5

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaPipeEntity_Fluid.java
+++ b/src/main/java/gregtech/api/metatileentity/implementations/GT_MetaPipeEntity_Fluid.java
@@ -263,8 +263,9 @@ public class GT_MetaPipeEntity_Fluid extends MetaPipeEntity {
             }
 
             if (mLastReceivedFrom == oLastReceivedFrom) {
-                ConcurrentHashMap<IFluidHandler, ForgeDirection> tTanks = new ConcurrentHashMap<IFluidHandler, ForgeDirection>();
-                
+                ArrayList<IFluidHandler> tTanksList = new ArrayList<>();
+                ArrayList<ForgeDirection> tDirectionsList = new ArrayList<>();
+
                 for (byte tSide = 0, uSide = 0, i = 0, j = (byte) aBaseMetaTileEntity.getRandomNumber(6); i < 6; i++) {
                 	tSide = (byte) ((i + j) % 6);
                 	uSide = GT_Utility.getOppositeSide(tSide);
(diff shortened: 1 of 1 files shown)


Prewf:
please make sure this fix works with multiple fluids actually going through the pipe. it looks like you only tested with just water going through all the pipes.

shpakovivan:
Two liquids flowed simultaneously:
mc6

Dream-Master:
@shpakovivan The problem after your change was the distillery accept no more fluids via normal pipes at the back side or the left and right side . Before it works well.

shpakovivan:
@Dream-Master Most likely the distillery returns incorrect information about the size of its tank (it can differ for different sides of the block).
I will fix pipes behavior. May be I fix distillery too.

Dream-Master:
Ok nice. If its done will try your fix again @shpakovivan

shpakovivan:
@Dream-Master I fixed the pipes.
I did not fix the "recipe-generated" machines. It's too "expensive".
These machines (including distillery) do not return fluid tank info.

Dream-Master:
@shpakovivan but distillery works now ?

shpakovivan:
Yes

Dream-Master:
Ok will test it later

Kortako:
@shpakovivan This code still does not work properly.  A simple boiler setup with a few high pressure coal boilers completely broke after testing a build with these changes merged in.



The steam output pipe (on the left side) will drain up until the 3 pipes that are connected to the boilers, and those 3 pipes at the end will stay full.  The boilers are full.  The supertank right underneath me does not fill up at all, even though those boilers are still producing steam.

@Dream-Master I suggest you un-merge this commit until it is fixed and tested very well.

draknyte1:
Why anyone ever merged the GT6 pipe/cable code is beyond me. 

Nobody had properly tested it, let alone reviewed it. So it's been a broken feature since merge to both branches.

Dream-Master:
undoing it and wait until it is fixed

Blood-Asp:
Yes, it can always be reverted if not stable until Release.
But If it was'n merged in The first place, almost no one would have looked at it and we likely would still not even know the issues exist.

shpakovivan:
mc1
@Kortako I'm not sure in your test. This pull request now contains are two commits. May be you have previously pulled the first only.

Dream-Master:
@shpakovivan after i pull it you dont add more code so far i can see. I did the pull with your lastest aviable code. Maybe in GTNH is something different then normal gt branch. Did you try it in GTNH modpack ?

shpakovivan:
I'm tested it on 5.09.32pre1 + some PRs

Kortako:
@shpakovivan I can test it again later tonight and see if I can give you more info to reproduce it.  I know that it didn't happen immediately, the tank would fill up for a little while and then it would completely break down after about 10 minutes I believe.

Edit: And yes, looking at the commit history on the GTNH branch, @Dream-Master did pull in both changes, so the bug is still present.

Kortako:
Ok, I might have more information to help reproduce this bug.  It seems like once a pipe is full and no longer flowing, it stops updating permanantly?

Here's what I did to make it happen.
1. Set up the same setup as I posted above.  Put charcoal in boilers, let it work for a while.  It looks like it works.
2. Attach a Infinite Fluid Tank filled with steam to the end, to make all the pipes fill up.  Make sure you use a pump that has greater throughput than the pipe.



3. After a little while, the pipes will fill completely, and stop sloshing.
4.  You should notice that the quantum tank does not fill up anymore.  It's stuck.
5.  If you break the quantum tank and replace it, it will start to fill up again.  But it will only drain the first two pipes, the pipes attached to the boilers will not drain.  See pictures below:





There are many other ways to make the pipes stop working, just by playing around with it.  It seems like an issue when a pipe is completely full, and possibly only when it has more than 1 "producer" input attached to it.

Hope that makes sense, I can do more tests if you need me to.

shpakovivan:
I tried to repeat the bug in different ways. Including pipe with "Infinite Steam Creative" block.
Pipes work correctly, But I do not use GT++ mod or other mods that change "internal behavior" of GT Unofficial.

I have a suspicion about the GT++ mod. It contains "New Custom Fluid Pipes for GT to suit Nuclear applications" for example.
@Kortako Can you exclude GT++ for test?

draknyte1:
My pipes use GT code to construct. So, it's nothing to do with my mod. 

No 'internal' behaviour is changed within GT by default by any mod other than GT itself.

shpakovivan:
@draknyte1 Then the reason is not in GT ++. Will seek

Kortako:
Alright, more tests.  It still doesn't work for me.  I'm really surprised you can't reproduce this, because it pretty much instantly breaks in every case I've tried.  For reference, I stripped down Minecraft to a minimal amount of mods; here's the list:
mods
So pretty much just IC2 and Gregtech.  And just to clarify, that build of GT is _your_ build directly from your fluid_pipes branch of your fork, not the version that was merged into GTNH.  The only thing I had to change is a dependency upgrade of COFHCore to 3.1.4-329, as it wouldn't build otherwise as it couldn't find the old version.

With only GT, setting up the same setup was just a little more annoying.  But I just used quantum tanks filled with water and steam to fill up the pipes real fast.


Below is a picture of the pipe right next to the quantum tank of Steam.  It's full.



And here's the pipe right next to it, constantly at a low value.  The full pipe next to it that I just showed above is not emptying into this pipe.



In this scenario, the Quantum tank on the very right (receiving tank) is only receiving steam from the 2 boilers on the right.  The left most pipe is full and stuck, so neither the leftmost boiler nor the left quantum tank will empty at all.

Can anyone else confirm that they get similar results to me?

shpakovivan:
I have two news: good and bad.
**I'll start with a good:** This is not my bug. Pipes work perfectly.

**Bad news:** We have two types of steam. They are incompatible.

First steam (generated with boilers) with metadata 4835:33 and temperature 375K
mc2

Second steam (received from IC steam cell in creative mode for example) with metadata 4835:12 and temperature 420K
mc1

One pipe block (or Quantum Tank) can be filled only with one type of steam. Two connected pipe blocks, filled with different steam, can not exchange fluid.
@Kortako You can repeat this effect on the current release of GT Unofficial (without my changes).

Kortako:
@shpakovivan Ah, that makes sense, although it concerns me.  It mostly means I need to do more tests, paying attention to what types of steam I'm using.  I still think _a bug_ exists somewhere, maybe in the GTNH branch with your changes merged in, because I deployed the GTNH version to an existing server instance and it broke the pipe system of one of my users with a similar very simple boiler setup (without the creative tank, of course).  That's how I found it in the first place, mostly by accident.  I'll have to see if I can reproduce it in another way.

shpakovivan:
In this case I can say, that you was realy can observe unaccustomed filling of pipes, because I changed principes of fluid distribution. You may need to change the pipeline layout.

Dream-Master:
@shpakovivan @Kortako  is it a bug or a feature?

shpakovivan:
@Dream-Master 
If you mean two incompatible types of steam - it is a feature (from past).
if you mean new principes of fluid distribution - it is a goal of changes :)

Pipes work well if you have one type of steam in pipeline. With two types of steam, pipeline are stop fluid exchange.

Dream-Master:
thats true.

Kortako:
I can't seem to reproduce what happened on my server anymore in SP, now that I know the difference between the two steams.  I still have the build of GTNH Gregtech with the change merged in, so I will test that again on my server for a while and see if anyone complains.  If no one notices anything for a few days, it may just be a mistake on my part.

Blood-Asp:
OK, question now is: Ready to merge or still WiP?

Dream-Master:
@Kortako @shpakovivan

shpakovivan:
I'm ready.

PR #1379, opened 2018-01-19
[1.12.2] World generators

Antifluxfield:
All generators are generally finished.
Tested for a few times, and no obvious issue found.


Exidex:
We have new repo -> https://github.com/Gregtech-CE/Gregtech
EDIT: You should probably log in to Discord

PR #1390, opened 2018-02-12
Fix the random maintenance issue of Fusion Reactor

Antifluxfield:
Fusion Reactors don't have a maintenance hatch, so it shouldn't occur such maintenance issue either.


Blood-Asp:
Did that really ever happen? I had fusion reactors running for hundreds of hours and never had an maintainance issue. And also no report about one...

Blood-Asp:
Took me a moment to understand how i did that years ago... Maintainance is cleared every multiblock check so it does at best sometimes happen for a few seconds... Better way to fix that is not just never call the damage method...

PR #1398, opened 2018-02-16
Gasoline revision

JohannesGaessler:
Currently Gasoline recipes use Acetone, which is only obtainable when using complicated chemistry.
They now use AceticAcid for simple chemistry.

The recipe for High-Octane Gasoline currently has an output FluidStack size of 32000L, which exceeds the 8000L output tank size of the Mixer.
To fix this the recipe now outputs a stack of Cells rather than a FluidStack.
Since this reduces the number of available inputs from five to four Octane is no longer required for producing GasolinePremium.
I think that this change will also make Gasoline production less rigid since the only way to obtain Octane was by cracking LightFuel.
The Octane Material has been removed since it no longer has any use.

Currently Gasoline requires the following five starting Materials: Naphtha, Refinery Gas, Methanol, Acetone/Acetic Acid, and Toluene.
I changed the recipes for Gasoline so that it no longer requires Gas.
I think that having five different starting materials makes the setup you'd need for producing Gasoline too complicated when compared to the reward.
Since this makes the Raw Gasoline intermediary material obsolete it has been removed.


jaskarth:
I agree with these changes

Blood-Asp:
ok, so thats what was missing...

Issue #1396, opened 2018-02-13
[5.09.32pre3] Recipe missing 


Baldrvolsk:
The new recipes do not work: the electrolyzer does not take CarbonDioxide and SulfurDioxide through the pipes #1307

Pilad:
I tested this. Work fine. gregtech-5.09.32pre3 + industrialcraft-2-2.2.828-experimental.
https://i.imgur.com/2dC3ZHT.png
https://i.imgur.com/IYTTCjv.png

Pilad:
And sulfur dioxide
https://i.imgur.com/cBWG7qV.png
https://i.imgur.com/laWQYcB.png

Pilad:
Recipes - Hide 
In GregTech.cfg --> B:MoreComplicatedChemicalRecipes = true
Set this option to false.

SapientMC:
@Pilad recipe should work with MoreComplicatedChemicalRecipes = true too. There are may be some difference in recipes, but it should work.

PR #1401, opened 2018-02-18
Fix Electrolysis of SO2/CO2

JohannesGaessler:
Since both SO2 and CO2 can be recovered as byproducts from the EBF they have been given special Electrolysis recipes with an increased EU cost.
When merging I incorrectly placed those recipes so that they're only added when using simple chemistry.
This PR fixes this mistake so that the recipes are added independently of the config that sets recipe complexity.

Related issue: https://github.com/Blood-Asp/GT5-Unofficial/issues/1396


NNM42:
cool. thx!

PiatraAlatyr:
The fix works correctly.

draknyte1 (approved):

Issue #1402, opened 2018-02-21
ODR 


NNM42:
https://github.com/Blood-Asp/GT5-Unofficial/blob/45342f945e9eb889fe4b0e0c624379a57427eb8e/src/main/java/gregtech/common/tileentities/machines/multi/GT_MetaTileEntity_OilDrillBase.java#L117this this code get error data for ChX<0 and\or ChZ<0.
For positive ChX and ChZ work correctly. ODR buggy now.

NNM42:
hm... after reload a game (map) fluid.amount in chunk to set new count. exp; before 301L after 244L. Maybe   because +-25% random.

PR #1410, opened 2018-03-06
Oil Drilling Rig fixes

Wo0kiee:
Fixed:

- ODR may get wrong location of chunks with negative and zero coords. #1402 

- ODR calculate wrong pump amount.

- ODR doesn't lift up pipes from empty oilfield, continue pumping 0L and consume energy.


draknyte1 (approved):

UhoMazino:
Should exclude GT_Log class because it was used for debug only and no need here.

PR #1419, opened 2018-03-16
Adv seismic prospector reworked

Wo0kiee:
Advanced Seismic Prospector is more a late game so radius expanded and because of that reduced amount of books 
Now it prospects 324(9 oilfields) oil chunks and gives min-max amount of whole oilfield(6x6).

image
image


Dream-Master:
Exactly what we need for GTNH :)

PR #1429, opened 2018-03-31
Make it support MOOORE EU

Technus:


draknyte1:
Don't care - https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/CONTRIBUTING.md.

Exidex:
Does it actually do anything?

Technus:
dis something new

PR #1430, opened 2018-03-31
Up cast total EU to long

Technus:
Fixes integer overflow for more demanding recipes by up casting to long. Only visual change in nei rendering.

Tested in GTNH fork.

Version for reference: 5.09.31


draknyte1 (approved):
Don't know why you didn't just edit the previous one.. but ok

Technus:
cos i realisze i needed to edit more than one file, and that is easier with just using fork

PR #1449, opened 2018-04-30
Add config for Small Boiler Automation

redmage17:
Default is off

### DIFF
--- a/src/main/java/gregtech/GT_Mod.java
+++ b/src/main/java/gregtech/GT_Mod.java
@@ -221,7 +221,7 @@ public class GT_Mod implements IGT_Mod {
         gregtechproxy.mFlintChance = tMainConfig.get(aTextGeneral, "FlintAndSteelChance", 30).getInt(30);
         gregtechproxy.mItemDespawnTime = tMainConfig.get(aTextGeneral, "ItemDespawnTime", 6000).getInt(6000);
         gregtechproxy.mNerfStorageBlocks = tMainConfig.get(aTextGeneral,"NerfStorageBlocks",true).getBoolean(true);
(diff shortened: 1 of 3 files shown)


draknyte1 (changes requested):
Your fucking Boolean logic is backwards, No.

draknyte1 (changes requested):
I guess you don't understand how booleans fucking work dude.

Blood-Asp:
I guess i have to give a hint:
```
if(value) return true
else return false 
```

is the same as 

`return value`

the other stuff was fine.

draknyte1:
You didn't need to give a hint because his PR is already fine?
Learn to read fucktard.

MauveCloud:
@draknyte1 did you seriously just insult Blood Asp (with very crude language I might add), who afaik is still in charge of GT5u?  redmage17 had already fixed the logic problem by the time Blood Asp saw it, in case you hadn't noticed.

Blood-Asp:
well yes, on mobile that was the only issue i saw but not that there was another push beween.
so that is just another thing to fix. works this way but should better be cleaned up.

and sry, that level of crude language? all you will get that way is more hate and certainly not improved code...

draknyte1:
@MauveCloud, your opinion is moot. 
@Blood-Asp learn to page refresh. Then you don't need to give useless opinions which make you look stupid. 

This PR was discussed and resolved outside of GitHub with Red and the final result can only be improved by simply returning the variable, as opposed to checking and returning a seperate Boolean. The JVM probably inlines this during compilation regardless.

draknyte1 (approved):

PR #1474, opened 2018-07-02
added range for bartimaeusnek

bartimaeusnek:
would be great if i could get this range for my mod crops++, my fork of EMT, and some other small addons


jaskarth (changes requested):
You should use 12501 - 13000

draknyte1:
I agree with Code, although curious as to why you skipped that range?

bartimaeusnek:
but that range is already reserved...

jaskarth:
It's for whoever asks next. In this case, it's you.

jaskarth (approved):

draknyte1 (approved):

PR #1496, opened 2018-09-18
Fix maintenance cover

UhoMazino:
Fix bug in maintenance cover logic when its try calculate turbine rotor needed maintenance by wrong and unpredicted ways.

### DIFF
--- a/src/main/java/gregtech/common/covers/GT_Cover_NeedMaintainance.java
+++ b/src/main/java/gregtech/common/covers/GT_Cover_NeedMaintainance.java
@@ -11,6 +11,7 @@ import gregtech.common.items.GT_MetaGenerated_Tool_01;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.fluids.Fluid;
+import gregtech.common.tileentities.machines.multi.GT_MetaTileEntity_LargeTurbine;
 
 public class GT_Cover_NeedMaintainance extends GT_CoverBehavior {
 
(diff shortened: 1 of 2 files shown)


draknyte1 (changes requested):
This change would break my GT++ Turbines which don't extend `GT_MetaTileEntity_LargeTurbine`.

There is no reason to make this cover strictly for Large Turbines either.. Other Multis exist which use the GUI slot.

NNM42:
what it mean "my gt++ turbines"?

SapientMC:
@draknyte1 Is it really the GT that must remain compatible with all the addons to it, and not vice versa?

draknyte1:
@Sapientmc, In this case, yes. 

You're turning compatibility with all mods to only supporting the GT turbine. Nothing mods can do to add support into the cover, so compatibility remains on GT.

SapientMC:
@draknyte1 but only GT turbine has rotors that takes damage and should be replaced on last points of durability.

mitchej123:
Make it work with both.  Previous behavior if not turbine and this behavior if it is a turbine

draknyte1:
There's no need to hard instance it as a GT turbine though, which is my point. Because the new field isn't necessary either..

Which means that my turbines are supported still...

NNM42:
@draknyte1, yep this is your point.
https://imgur.com/ECwyOdb

draknyte1:
@NMM42, either make the changes I've requested or I'll just close your PR. Take your smartass attitude elsewhere, because nobody else will approve it. 🤔

Issue #1456, opened 2018-05-16
Nettlesome Phosphor Name


YoungGT:
There‘re four materials that contain _phosphorus_ : **phosphorus** ,**phosphor** ,**phosphate** and **phosphorous pentoxide**. The last one is very identifiable ,but those three items are very indecipherable in GT. But why? The reason is that the names of them in GT are confused. Let us check them out.
I list them up![](https://user-images.githubusercontent.com/30253037/40097235-96346c42-5907-11e8-9d02-1ed005ea7c20.png)
And we can find that the names of **phosphorus** and **phosphor** wrong easily.
I'm trying to translate the language file into Chinese and it really makes me vexed. So I'd like to correct the primary names by the containing elements.

draknyte1:
As Dream has found, this breaks mods using these materials. The old ones should be deprecated, not removed. Look at other examples within the materials clas.

draknyte1:
@Blood-Asp I would’ve noticed this upon review, your reckless merging is a pain in the ass.

PR #1503, opened 2018-12-09
Fix #1439 and #1456

### LINKED ISSUE #1439: Galena ore processing balance bug
Galena Dust can be processed several ways.
* 1 Galena Dust -> 1 Lead Ingot (smelting)
* 1 Galena Dust -> 0.375 Lead Dust + 0.375 Silver Dust + 0.25 Sulfur Dust (electrolyzer)
* 1 Galena Dust -> 1.5 Lead Ingot + 1 Silver Ingot (primitive blast furnace)
* 1 Galena Dust -> 1.5 Lead Ingot + 0.67 Silver Ingot (blast furnace chemical smelting)

Compare the last 2. You can clearly see that the more primitive one gives more output. This is illogically considering the last one is performed at a higher tech level with higher efficiency and requires some decent work with gases to be performed. 

I suggest increasing silver yield from the last recipe from 6 nuggets to at least 9 nuggets, so it is equal to the primitive blast furnace recipe, and ideally the silver output should be 10-11 nuggets to make the energy and resources spent on performing the last recipe provide at least some profit!

YoungGT:
Change nuggets of the primitive blast furnace recipe to 6 and EBF to 9, should fix #1439.
And correct phosphorus's name according to #1456. But I just change Chinese and English language files, I can't speak Russian and German, hope somebody help me.


src/main/java/gregtech/loaders/postload/GT_CraftingRecipeLoader.java
```diff
@@ -490,7 +490,6 @@ public void run() {
         GT_ModHandler.addShapelessCraftingRecipe(new ItemStack(Items.gunpowder, 6), GT_ModHandler.RecipeBits.NOT_REMOVABLE, new Object[]{OrePrefixes.dust.get(Materials.Charcoal), 	OrePrefixes.dust.get(Materials.Charcoal), 	OrePrefixes.dust.get(Materials.Charcoal), 	OrePrefixes.dust.get(Materials.Sulfur), OrePrefixes.dust.get(Materials.Saltpeter), OrePrefixes.dust.get(Materials.Saltpeter)});
         GT_ModHandler.addShapelessCraftingRecipe(new ItemStack(Items.gunpowder, 6), GT_ModHandler.RecipeBits.NOT_REMOVABLE, new Object[]{OrePrefixes.dust.get(Materials.Carbon), 	OrePrefixes.dust.get(Materials.Carbon), 	OrePrefixes.dust.get(Materials.Carbon), 	OrePrefixes.dust.get(Materials.Sulfur), OrePrefixes.dust.get(Materials.Saltpeter), OrePrefixes.dust.get(Materials.Saltpeter)});
 
-        GT_ModHandler.addShapelessCraftingRecipe(GT_OreDictUnificator.get(OrePrefixes.dust, Materials.IndiumGalliumPhosphide, 3L), GT_ModHandler.RecipeBits.NOT_REMOVABLE, new Object[]{OrePrefixes.dust.get(Materials.Indium), OrePrefixes.dust.get(Materials.Gallium), OrePrefixes.dust.get(Materials.Phosphor)});
```
Blood-Asp, line 495: Why did this recipe get removed without replacement?

Blood-Asp:
Looks mostly fine besides the one removed recipe.

Changing material names will break minetweaker recipes but should not be that bad. As materials are saved as IDs and not by their names renaming them should not remove them form the saves like it happens with fluids. 

No idea if the oredict entrys are used by other mods. can't remember any that do.

YoungGT:
Well that recipe should be a mistake, but as for mt, don't know and can't fix thought....

draknyte1:
https://github.com/Blood-Asp/GT5-Unofficial/issues/1456#issuecomment-446327923

Because you’re a useless dev, what about explicit calls to those materials? That will break.

Blood-Asp:
And how about you shut up. If everything you write is an insult it is not needed whatevery quality on the coding side.

draknyte1:
f866e385-9c36-4689-a80d-d94a1e43427e
But I wasn’t wrong though. 🤔

> And how about you shut up. If everything you write is an insult it is not needed whatevery quality on the coding side.

Didn't actually respond to my statement about you being wrong...
Revoking my git privs and muting me on all discord channels is a weak move to avoid any fault. 

> i made a check if there is any mod using that and could not find any, so i merged.

Bullshit. https://github.com/draknyte1/GTplusplus/blob/master/src/Java/gtPlusPlus/core/material/ELEMENT.java#L29

YoungGT:
OK I know it's such serious a problem that caused by me because of my poor acquaintance of Java and GT5U. I'm so sorry about it. I won't leave it, I would fix on today's evening(my evening here). 
And please don't fight, please.

Blood-Asp:
It's not a serious problem. should there be some mod using these materials they have to update their materials too, but no idea if there even one exists. otherwise i would not have merged it.

I have no problem with pointing out issues but the proper way would be to write how to do it better instead of stupidly repeating something is wrong and insulting everyone.

PR #1504, opened 2018-12-12
Additional input bus mode

Dimach:
Add disable sort option for input bus. Default behavior will stay same
Disable sort will disable automatic item transfer to lower slots


draknyte1 (approved):

draknyte1 (changes requested):
In hindsight, this could use minor adjustments.

1) Tool-tip could be more clear. Without knowing, 'Sort mode on/off' is very nondescript.
2) Why does it shuffle items if disableSort? You'd expect it to do NOTHING...
3) onScrewdriverRightClick is way too complex for what you're doing.. The first if is 100% pointless.
4) I don't think we should even refer to it as sorting. We probably should think up a better term.

Discord community has agreed upon `Automatic Item Shuffling` instead of referring to it as `sorting mode`. Which resolves points 1 & 4.

draknyte1 (approved):

PR #1505, opened 2018-12-12
Deprecate the old materials that were removed

YoungGT:
Deprecate the old materials that were removed, shouldn't break mods using these materials.
And the name of tricalcium phosphate is the same as phosphorus's, so I add "NEW" to phosphorus to distinguish them.


Blood-Asp:
Thats the wrong way to fix it. 
As example for deprication look here: https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/api/enums/Materials.java#L758

Blood-Asp:
Also pull the main repo changes. otherwise i have to resolve merge conflicts.

YoungGT:
well too late for me....

Blood-Asp:
Did it myself to get an end to this: https://github.com/Blood-Asp/GT5-Unofficial/commit/88a7715fcaa8417f15713175de3b5f5a45f65266

PR #1508, opened 2019-01-11
added my addon (bartworks) to the readme

bartimaeusnek:


draknyte1 (approved):
Can you adjust the ID allocations for Tec? It lists two different names for the same person.

PR #1511, opened 2019-01-16
Port digital transformers by Spartak

Dimach:
Add digital transformers - block that allow adjust its in/out voltage (from ULV to transformer tier) and amperage (0-64). 
In game they called "Adjustable transformers" and have various tier (EV-MAX).


draknyte1:
> Uses someones elses ID range.
> Ridiculously OP for standard GT.

No from me.

Blood-Asp:
The concept itself sounds interesting and the IDs can be changed but the recipes are certainly much too cheap. Will likely merge soon and readjust.

SapientMC:
> recipes are certainly much too cheap.

Actually no. It's require circuits of next tier than usual transformers, so it's more expensive than chain of existing transformers.

Blood-Asp:
balancing is not just adding up the costs of the components. here you have the ablility of many transformers combined into a single one in one block. 
the usefullness is more like quadric instead of linear so should be the costs.

Maybe: 2 transformer tiers in one = 1 circuit tier higher, 3 transformer tiers in one 2 circuit tiers higher, ect...?

SapientMC:
> 3 transformer tiers in one 2 circuit tiers higher, ect...?

Than it will not be usable at the same time as existing transformers and will require to replace trans-chains with Adjustable one much later.

BTW I've already tested 2k and 8k variants in real game. And it needs very long time to craft them even via full automated ME-Network.
Try it in game by yourself.

PR #1525, opened 2019-03-23
Add usage for Tantalum №2

Spartak1997:
Add Tantalum foil special for SMD Capacitor recipes


draknyte1 (changes requested):
doesn’t actually give any details what this does.

Spartak1997:
This PR need for previous addition. Because in this thing I add ordict for Tantalum foil and plate special for recipes

draknyte1:
https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/CONTRIBUTING.md
So make them one PR and close this..

Blood-Asp:
i too would have preferred to have them in one PR, but not waiting for that...

PR #1527, opened 2019-04-19
Improved LuV+conveyor modules

AndreySolodovnikov:
now LuV+ Conveyors can work with more than 1 stack per tick: luv: 2, zpm:3,uv:4;


draknyte1 (changes requested):
You can’t change the constructor, Add a new one.

AndreySolodovnikov:
added new constructor =)

draknyte1 (approved):

PR #1528, opened 2019-04-19
Added an Advanced Processing Array

AndreySolodovnikov:
added improved versing of basic processing array, it has it own casings and is larger(5x5x5). Advanced processing array can handle up to 64 machines


draknyte1 (changes requested):
Doesn’t need a config option. 
Doesn’t need a new block class for the casings. (There’s several free spaces in the existing classes, just be aware texture IDs > 63 on page 0 are reserved.

src/main/java/gregtech/common/blocks/GT_Block_Casings8.java
```diff
@@ -13,23 +13,28 @@
...
+        for (int i = 0; i < 2; i = (i + 1)) {
             Textures.BlockIcons.casingTexturePages[1][i+48] = new GT_CopiedBlockTexture(this, 6, i);
         }
         GT_LanguageManager.addStringLocalization(getUnlocalizedName() + ".0.name", "Chemically Inert Machine Casing");
         GT_LanguageManager.addStringLocalization(getUnlocalizedName() + ".1.name", "PTFE Pipe Casing");
+        GT_LanguageManager.addStringLocalization(getUnlocalizedName() + ".0.name", "Robust HSS-G Machine Casing");
 
         ItemList.Casing_Chemically_Inert.set(new ItemStack(this, 1, 0));
         ItemList.Casing_Pipe_Polytetrafluoroethylene.set(new ItemStack(this, 1, 1));
+        ItemList.Casing_RobustHSSG.set(new ItemStack(this,1,0));
```
draknyte1: Didn’t set the meta.

AndreySolodovnikov:
Updated. now everything should work

PR #1529, opened 2019-04-24
Added some fixes for old bugs

AndreySolodovnikov:
This PR includes 
1.fix for HV and LV circuits that cannon be put in any machine
2.fix for assembly line books and data sticks on server
3,fix for automaintance hatch. Now it can hold only 1 stack of each item that it needs for maintating


Blood-Asp:
hmm, after so long some work done again.

guess i have to find some free time to check PRs soon again

draknyte1 (changes requested):
AL server stuff is not required. Consult the GTNH fork. 
Hard coding scanner recipes for GC blueprints is also a no. 

Will review rest later today.

SapientMC:
@draknyte1 Could you express in more detail?
Our level of English does not allow us to understand exactly what is wrong.

Blood-Asp:
@SapientMC pretty sure he means: GT New Horizon has already done a fix for the Assemblyline on servers that works without all that duplicate code. checking out how it is done there and porting it over here is the preferred way.
And i fully agree that seperated client/server code for machines is a bad idea.

The part about the Galacticraft blueprints should be on point and clear...

AndreySolodovnikov:
@draknyte1 NH fork has almost same fix for AL books. Should I just use previous commit, without ALServer stuff? https://github.com/GTNewHorizons/GT5-Unofficial/commit/3e5b4f63967ed306a74f8f597a66185d76c7f335#diff-8e7d8a5ff9ea320069c1742efa0220b2 this commit adds ALServer . Or should I port only this commit https://github.com/GTNewHorizons/GT5-Unofficial/commit/8ce5ad6bd06413f3c9148cd586fef03a17fb920d#diff-8e7d8a5ff9ea320069c1742efa0220b2

draknyte1:
Port whatever they did, it works, it’s tested..

AndreySolodovnikov:
Well I've ported that.. and I copied GT-Scanner class accidently copied GC patterns part, I can delete it, then my PR will become port of GTNH AL fix

AndreySolodovnikov:
Updated, deleted GC patterns. now AL fix is same with this commit https://github.com/GTNewHorizons/GT5-Unofficial/commit/3e5b4f63967ed306a74f8f597a66185d76c7f335#diff-8e7d8a5ff9ea320069c1742efa0220b2

PR #1532, opened 2019-04-29
Screen cover pr

AndreySolodovnikov:
added new more industrial like recipe for screen cover


draknyte1:
Please create a new fork whenever you create a new PR. 
We do NOT need the same commit history across several merges.

draknyte1 (approved):

PR #1533, opened 2019-05-18
Added zincite  Carbothermic Reduction recipe to electric blast furnace

AndreySolodovnikov:


draknyte1 (commented):
Zincite has no uses or any way to be obtained, right?

draknyte1 (changes requested):
Please explain why this should be merged..

AndreySolodovnikov:
Zincite can be obtained with sphalerite  oxidation recipe like other material (cobalt for example) but it has no carbotermic reduction recipe, only electolizer. So if there is no c.r. recipe oxidation recipe is useless and best way to process sphalerite is use pritive blast furnace

PR #1534, opened 2019-05-18
added hard radon recipe and config for it

AndreySolodovnikov:


draknyte1 (changes requested):
This seems utterly pointless and very specific. 

Please explain why this should be merged.

AndreySolodovnikov:
This configuration is useful for players ( and modpacks) who like good balanced game experience. Without this configuration you can use large chemical reactor and get tons of radon not using much energy( fusion reactor recipe is more expensive so it is useless)

PR #1538, opened 2019-06-25
NaOH electrolyzis recipe now is more expensive

AndreySolodovnikov:
Sodium is used for cheaper water decomposition, but due to this recipe Hydrogen is too cheap. You can use LCR on 2048V + 4 MV electrolyzers and get supercheap Hydrogen that will be profitable when used as fuel in Gas Turbine(this combination gives you ~1000 eu per tick)


SapientMC:
To prevent some questions. We should not have an endless source of energy from nothing cheaper than solar panels. Hydrogen with the old recipe was just that.

draknyte1:
1000eu/t is pathetic free energy.

PR #1541, opened 2019-09-04
Added BartWorks integration for BasicMachine recipes

bartimaeusnek:
this changes EV+ machines to use my special kind of glass instead of regular glass.


draknyte1 (approved):
Can you put the swich in {}?
It looks out of place from the diff.

draknyte1 (approved):

PR #1546, opened 2019-11-16
fix #1545 GT_MachineRecipeLoader run() method too big

### LINKED ISSUE #1545: GT_MachineRecipeLoader run() method too big
In the `Unstable` branch, the [`loaders/postload/GT_MachineRecipeLoader.java/run()`](
https://github.com/Blood-Asp/GT5-Unofficial/blob/unstable/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java#L56) method is too big.

> ```none
> src/GT5-Unofficial-blood/build/sources/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java:56: error: code too large
>     public void run() {
>                 ^
> ```

leagris:
Move excess code from `run()` to the chained `run2()`

Related issue: https://github.com/Blood-Asp/GT5-Unofficial/issues/1545

### DIFF
--- a/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java
+++ b/src/main/java/gregtech/loaders/postload/GT_MachineRecipeLoader.java
@@ -29,30 +29,30 @@ import java.util.Map;
 
 public class GT_MachineRecipeLoader implements Runnable {
     private final MaterialStack[][] mAlloySmelterList = {
-    		{new MaterialStack(Materials.Tetrahedrite, 3L), new MaterialStack(Materials.Tin, 1L), new MaterialStack(Materials.Bronze, 3L)}, 
-    		{new MaterialStack(Materials.Tetrahedrite, 3L), new MaterialStack(Materials.Zinc, 1L), new MaterialStack(Materials.Brass, 3L)}, 
-    		{new MaterialStack(Materials.Copper, 3L), new MaterialStack(Materials.Tin, 1L), new MaterialStack(Materials.Bronze, 4L)}, 
-    		{new MaterialStack(Materials.Copper, 3L), new MaterialStack(Materials.Zinc, 1L), new MaterialStack(Materials.Brass, 4L)}, 
-    		{new MaterialStack(Materials.Copper, 1L), new MaterialStack(Materials.Nickel, 1L), new MaterialStack(Materials.Cupronickel, 2L)}, 
-    		{new MaterialStack(Materials.Copper, 1L), new MaterialStack(Materials.Redstone, 4L), new MaterialStack(Materials.RedAlloy, 1L)}, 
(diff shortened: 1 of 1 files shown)


draknyte1 (changes requested):
Eclipse and ForgeGradle have no issue with this..
Not approved, pending review.

leagris:
What are the requested changes?

draknyte1:
Nothing. I just need to verify this is actually required. 
I’ve got six open PRs and none had this issue in eclipse nor ForgeGradle. 
Which means the unstable branch compiles fine..

leagris:
Just for info I had the error using gradle build, and OpenJDK8 in a Linux environment.
There is possibly a tiny opcode difference. Although this also reveal how the run method is close enough to the maximum allowed size that it cannot compile on tiny compiler variations. This patch moves a few lines of code to the run2 method that was created there earlier for this exact purpose of offloading the already crowded run method.

draknyte1:
> OpenJDK8

You and literally 95% of everyone else still using Java 8, lol. 

I can obviously see what your change does, It’s just the lowest priority PR currently open and will be reviewed and merged last.

MauveCloud:
I did a little searching related to this, and discovered that Eclipse has its own built-in Java compiler.  I was not aware of this, since I don't use Eclipse (and probably could have missed this detail even if I did).  For all I know, said compiler could be splitting up the method "under the hood".

However, even with this change, the run() method is still alarmingly long, and could easily be split into numerous smaller methods (e.g. one for adding circuit-related recipes, one for adding assembly line recipes, one for adding vacuum freezer recipes, etc.).  It might also be possible to read these recipes from a file (either external or embedded in the mod jar as an asset), but I'm doubtful that's worth the extra effort.

leagris:
`GT_MachineRecipeLoader` has been subverted and abused since very long ago.

It was initially designed to load Machine recipes with ingredients not having Ore Dictionary entries and no Unification rules, like Torches (no OreName) with stick and IC2 Harz (No OreName).

This whole class is an ever growing spaghetti growing with each contribution.

This patch here is not meant to refactor the whole mess, but to allow at least compilable code as-is.

If someone wants to really work in this refactor. I'd prioritize adding true OreDictionary and Ingredient Alternatives to all Gregtech Machines.

Then probably split this class into one per mod integration.

Then probably implement a ingredient definition class to the model of MineTweaker's/

This would save huge load times.
This would save so many NEI pages displaying recipes of same output but many different item alternatives.

Ubus99:
> Nothing. I just need to verify this is actually required.
> I’ve got six open PRs and none had this issue in eclipse nor ForgeGradle.
> Which means the unstable branch compiles fine..

I have the same problem with IntelliJ's build in gradle, and gradlew doesn't work either.

PR #1548, opened 2019-12-13
Fixed Power injection for Multiblocks.

draknyte1:
All Energy Producing Multis will explode if they cannot handle the output load.
All Energy Producing Multis also now support multi amp dynamo hatches, which may be added by addons.


Dream-Master:
@draknyte1 is this code for GTNH GT too ?

draknyte1:
No. 

It *Is* the NH code. 
Which is a slightly modified version of my original code.

Dream-Master:
ok

Dream-Master:
ok

PR #1551, opened 2019-12-13
Added a better way for the Processing Array to handle recipes.

draknyte1:
Now addons can register their machines for PA support.


src/main/java/gregtech/api/util/GT_ProcessingArray_Manager.java
```diff
@@ -0,0 +1,29 @@
+package gregtech.api.util;
+
+import java.util.HashMap;
+
+import gregtech.api.util.GT_Recipe.GT_Recipe_Map;
+
+public class GT_ProcessingArray_Manager {
+
+	private static final HashMap<Integer, String> mMetaKeyMap = new HashMap<Integer, String>();
+	private static final HashMap<String, GT_Recipe_Map> mRecipeCache = new HashMap<String, GT_Recipe_Map>();
```
draknyte1, line 10: Should probably change this to a set type collection or check for existing key to save a cpu cycle or two.

PR #1572, opened 2020-10-19
Multiblock hatch changes

MauveCloud:
Input hatches no longer allow external draining (via pump cover, ME fluid import bus, etc.), and can only be filled from the side with the opening.
Output hatches no longer allow external filling (via ME fluid export bus, etc.), and can only be drained from the side with the opening.
Multiblocks are still able to populate output hatches with fluid (I personally tested this with a distillation tower and oil as input).
Both hatch types still allow both filling and emptying of fluid containers via GUI.


MauveCloud:
Cancelling this pull request after some clarifications in #1544.

PR #1575, opened 2020-10-27
Fluid Input and Output in single-block machines

MauveCloud:
Added gui fluid input/output: It is now possible to click with a fluid container on the internal fluid tanks in the gui of a basic (single-block) machine to add or remove fluid (depending on whether the container is empty).

Added option to make machine input tanks less forgiving: Defaults to false, but if set true, players will not be able to simply use an empty fluid container to recover fluids mistakenly added to the input tank of a machine.


leagris:
Awesome change!

Would you be able to support variable content containers like GT Steel Cells,  Universal IC2 Cells, GT++ Adjustable flasks, OpenBlock Tanks, Portable Tanks, Drums... ?

MauveCloud:
I adapted code from GT_MetaTileEntity_BasicTank, so any container that will work in an input/output hatch or quantum tank should also work here - I tested with IC2 cells, universal fluid cells, GT steel cells, and volumetric flasks (which are part of main GT now - edit: I notice GT++ has "large" and "gigantic" volumetric flasks, but aside from having larger default capacities, I'm not sure what advantage they offer).  Basically any container that implements IFluidContainerItem or an IC2 cell for which there is a filled version defined should work (though no partial filling/emptying).  I hadn't realized OpenBlock Tanks could be used as item containers, and drums (from Extra Utilities) might be too big.

MauveCloud:
Update: I ran into a snag with this.  It wasn't obvious at first when I was testing in creative mode, but when using a custom build in survival that included this change, I noticed that the stacks (of fluid containers) weren't updating immediately - I often had to exit the machine gui and then it would show the new stack sizes.  Calling detectAndSendChanges() on either that machine's container object or the player inventoryContainer was enough to update the filled/emptied containers added to inventory, but I haven't figured out how to make it immediately update the stack of on-cursor containers.

Edit: the key seems to be the return value.

Issue #1324, opened 2017-12-10
Items Registered to plateWood added to plankWood automatically.


draknyte1:
```
for meta in [32470, 32471, 32472, 32473, 32474, 32475, 32476, 32477, 32478, 32479, 32480, 32481, 32482, 32483, 32484, 32485, 32486, 32487, 32488, 32489, 32490, 32491, 32492, 32493, 32494, 32495, 32496, 32497, 32498, 32499] as int[] {
  <ore:plateWood>.add(<gregtech:gt.metaitem.02>.definition.makeStack(meta));
}
```
This puts everything into plankWood as well, which is not a desired effect.
![2017-12-10_14 21 02](https://user-images.githubusercontent.com/3060479/33801909-d8cdff3c-ddb5-11e7-81de-420a13664d2e.png)
![2017-12-10_14 21 14](https://user-images.githubusercontent.com/3060479/33801910-d9066e30-ddb5-11e7-8358-3439b4091757.png)

Issue #1573, opened 2020-10-20
[Suggestion] Specify time units in assembly line recipe books


MauveCloud:
I scanned a superconductor wire to get the recipe book for a fusion control computer, and the end of the first page says "Production time: 50".  Is that 50 ticks? 50 seconds? 50 minutes? 50 _hours_?  It would be nice to have that clear ahead of time and shown in the book.

Issue #1568, opened 2020-10-01
[Bug]The LV Energy Hatch and the Maintenance Hatch lost their texture


JustAlkaid:
v5.09.33.02, on the left is LV Energy Hatch with its back to me,, on the right is Maintenance Hatch. They were mounted on the cleanroom.
![QQ截图20201001215257](https://user-images.githubusercontent.com/72213408/94827204-bcf94880-043a-11eb-8bac-98f3144886e5.png)

Issue #1556, opened 2020-02-04
[5.09.32pre6] Advanced Circuit recipe annoyance


MauveCloud:
When I go to check the recipes in NEI for making an HV-tier circuit with an MV-tier circuit assembler (called a Processor Assembly, using a plastic circuit board) or an EV-tier circuit assembler (called a Nanoprocessor, using an epoxy circuit board), I have to scroll past the recipe for the IC2 Advanced Circuit in an LV-tier circuit assembler (which indirectly uses phenolic circuit boards) _every time_, and quickly gets annoying.

I've tried searching the source code (local copy, since searching GitHub often doesn't work as well), but I haven't had much luck figuring out why this happens.  Maybe somebody else will have better luck.

MauveCloud:
Okay, possible clue: I started a new world with Mekanism included in the modlist, and now the circuit assembler recipe that would normally make an IC2 advanced circuit instead makes a Mekanism Advanced Control Circuit, at least when looking up the recipe in NEI.  I haven't (yet) figured out why that happens either, but I'm guessing it's related.

MauveCloud:
It took a little digging, but this can be fixed with a one-line change:
https://github.com/Blood-Asp/GT5-Unofficial/blob/77a3c956e510b7c8ef4654054bcb16a0fccb86f7/src/main/java/gregtech/api/enums/OrePrefixes.java#L226
That first "true" in the argument list is for "aIsUnificatable", meaning that circuits can be oredict unified (though the GregTech-specific ones have subtags to prevent that).  I see no reason to unify circuits, and setting that false allows proper lookup of the Processor Assembly and Nanoprocessor (without having to scroll past the IC2 Advanced Circuit recipe).

PR #1578, opened 2020-10-28
Small fixes

MauveCloud:
This is a pull request for fixes that seemed too small for separate pull requests.  These can be evaluated and approved separately and manually added to the base if desired.

Details will mostly be in the commit messages, though for one I forgot to mention issue #1324


MauveCloud:
Apparently mentioning issues in the commit messages will only link them back to those commits, not to this PR, so here are the other affected issues:
#1555 
#1556 
#1573

MauveCloud:
Implemented #1553

MauveCloud:
Implemented #1554

draknyte1:
For 1553, just take the existing code from the GTNH repo. It’s *much* cleaner.

draknyte1:
Let me know when this is done and I’ll approve.

MauveCloud:
Chemical formula subscripts inspired by https://github.com/GTNewHorizons/GT5-Unofficial/pull/38 , but hopefully my version is cleaner than that attempt.  There might already be something similar in GT++ (in code from before it went private), but I didn't know where to look.  Edit: actually, I found the similar function in GT++ before the commit below, it helped inspire me.

MauveCloud:
Added a copy of https://github.com/GTNewHorizons/GT5-Unofficial/pull/331 for inclusion in main fork.

MauveCloud:
Should help with #1568 which as far as I can tell was caused by https://github.com/Blood-Asp/GT5-Unofficial/commit/f207774889893e60d034e17a3c468a431af9d1de

MauveCloud:
@draknyte1 I think this is basically done.  If I think of more "small fixes" or you want to list some changes from GTNH to carry over, I can make a new PR for those.

Issue #1576, opened 2020-10-27
[Suggestion] Make machines save state when harvested


MauveCloud:
Originally suggested by leagris in #1544 

> Then in a blink of questionable motivation, he added the overly complicated plunger to purge input tank whereas wrenching out the machine is always a simpler and less costly option. Because he expected to make the tilemachine NBT data to be persisted on harvest but failed the code. So machine inventory is not saved to the item in harvest and the plunger is made useless.

Looking at the changelog, https://mechaenetia.com/changelog/#Changelog I think you are slightly mistaken about the motivation:
> 5.02h
> ...
> Added Plunger, which can be used to clear a clogged GT Item Pipe System just by rightclicking any Pipe inside the System.

> 5.02j
> Added two new abilities to the Plunger: 1. Cleaning Fluid in Tanks and Pipes and 2. Cleaning Essentia Containers and Tubes from Essentia.

Admittedly, the item pipe system can alternately be cleared by connecting a chest.  I'm not sure about essentia containers and tubes.

As far as making the tilemachine NBT data persisted on harvest, the framework is already there:
https://github.com/Blood-Asp/GT5-Unofficial/blob/77a3c956e510b7c8ef4654054bcb16a0fccb86f7/src/main/java/gregtech/api/metatileentity/BaseMetaTileEntity.java#L1249
It's just that most GT tile entities don't do anything yet with that "setItemNBT" method.  A lot can be copied from the "saveNBTData" method, but there are a few details that probably shouldn't be saved when harvested:
1. Facings - the player might be planning to move the machine, and the new location might be the opposite orientation of the old one.
2. Progress - if the player insists on breaking the machine while it is running, it might be intentional to interrupt the operation.
3. Item inventory - is it really worth saving this, rather than letting the items drop if the player neglects to deal with them beforehand?

leagris:
> Item inventory - is it really worth saving this, rather than letting the items drop if the player neglects to deal with them beforehand?

Yes it be nice to keep molds, shapes, integrated circuits.

Notes about persisting machine covers and orientation:

Currently, machine front is determined when placing. This is absolutely ok to stay as-is.
But output side is always default rear, except for single-block generators who ave their output side at front.
Covers are persisted but their data is stored as absolute world side. I mean it would be much more practicable to have covers persisted relative to the front side, so as when placing a previously harvested machines, left, right, rear, front covers are still left, right, rear front relative to player, regardless of the new machine placement.
[...]

PR #1579, opened 2020-10-28
Save machine state

MauveCloud:
See #1576 
Allows machines to save more state than before when harvested (they already saved covers, paint, and some other things):
1. fluid in input and output tanks
2. fluid and item auto-output settings
3. stored energy
4. processing enabled/disabled state (i.e. whether you stopped it with a soft mallet)
5. item inventory
6. original main facing (so covers can be rotated for new main facing) - extra semi-related feature: rotating main facing (even if not harvesting machine) will rotate covers as well


draknyte1:
Machines should not keep their contents when broken.

MauveCloud:
I wonder how @leagris will react to that.  Personally, I don't insist on it, since there are other mods that provide ways to move machines while keeping inventory, such as Mekanism cardboard boxes and apparently the DartCraft Force Wrench (and possibly others that I'm less familiar with).  I guess I should trim this down to just the cover rotation part?

leagris:
I just react that dropping fluid inventory on harvest just make the plunger and similar devices totally useless.

PR #1592, opened 2021-11-12
Minor Fixes.

draknyte1:
Fixed IC2 IBoxable handler, GT Tools now go in IC2 Toolbox.
Made Toolbox work for maintenance issues.
Updated IC2 version.


draknyte1:
image
image
image

Soldering Iron works if it's not the NULL one with no charge, also works as NULL one in creative.

PR #1597, opened 2022-07-08
1st attempt on fixing small boilers automation. Seems to work correctly 

Vitium92:
See https://github.com/Invincible92/GT5-Unofficial/commit/f16a5dab181325e18d9dcda064c7ae8a5430f58b#commitcomment-78010299

The release version on the GT5 GitHub doesn't properly work when the config option "_B:AllowSmallBoilerAutomation_" is set to true. If set as such, it pulls from inventories as hoppers _but_ the combustible materials are put **into** the water buckets/cells section, and then, they get _**immediately**_ pulled from the hopper below the boiler, which it is not the correct way. 

To correct this, I have had to look into the GT++ code through IntelliJ IDEA decompiler (_I'm truly sorry for that @draknyte1_) and compare the boiler _.class_ files betweeen the two sources.

I have taken a screenshot of the change working as it was intended ("_B:AllowSmallBoilerAutomation_" set to true)

And this is with "_B:AllowSmallBoilerAutomation_" set to false.

Hope the change is OK to the PR.


draknyte1 (approved):
The code is publicly available under LGPLv3, so no problem.

